Issue #45, opened 2018-07-07
Add “Your Response” / 「答え」hint text to review prompts.


rcatalano:
One thing that I love about the WaniKani web app is that I only need to glance at the text field and instantly know if I’m supposed to enter English or Japanese. See screenshots for examples. With Tsurukame, I need to process either “Meaning” or “Reading” and then switch my brain to either English or Japanese mode. I know it doesn’t sound like a big difference, but I have a have a very hard time using the app without this cue.

![img_0059](https://user-images.githubusercontent.com/471944/42413147-14d5a0f8-81cf-11e8-8417-ae1648c41b81.jpeg)
![img_0060](https://user-images.githubusercontent.com/471944/42413158-40fdcf02-81cf-11e8-98f9-428ac9539262.jpeg)

PR #46, opened 2018-07-07
(#45) Adds “Your Response” / 「答え」hint text to review prompts.

rcatalano:
This resolves issue https://github.com/davidsansome/tsurukame/issues/45.

Now setting `answerField.placeholder` to either `@"答え"` or `@"Your Response"`, as appropriate.

### DIFF
--- a/ios/ReviewViewController.m
+++ b/ios/ReviewViewController.m
@@ -388,6 +388,7 @@ static UIColor *kDefaultButtonTintColor;
   NSString *taskTypePrompt;
   NSArray *promptGradient;
   UIColor *promptTextColor;
+  NSString *taskTypePlaceholder;
   
   switch (_activeTask.assignment.subjectType) {
     case TKMSubject_Type_Kanji:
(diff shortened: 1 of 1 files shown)


ios/ReviewViewController.m
```diff
@@ -426,23 +429,14 @@ - (void)randomTask {
   
   // Animate the text labels.
   UIViewAnimationOptions options = UIViewAnimationOptionTransitionCrossDissolve;
-  [UIView transitionWithView:self.successRateLabel duration:kAnimationDuration options:options animations:^{
+  [UIView animateWithDuration:kAnimationDuration delay:0.f options:options animations:^{
```
davidsansome: This doesn't actually work - try it out in the simulator with slow animations turned on!
The problem is that animateWithDuration is for changing animatable properties of views, and transitionWithView is changing the view hierarchy - adding or removing views.  I guess changing UILabel's text property actually removes the old text view and adds a new one?  So it needs to be done in a transition, not a property animation.

rcatalano: You're right, it looks better with the transition.

PR #50, opened 2018-08-07
ignore spaces in reading answers

wedgeV:
i sometimes hit the space bar by accident, inserting random spaces into the reading answer, this should strip out any space before doing the check.

this is assumes there is never any space required in any reading answers (not sure if that's the case)


davidsansome:
Thanks for the change!  Sorry for taking so long to merge it!

Issue #30, opened 2018-05-02
Review option: random fonts (like Jitai)


davidsansome:
It would be useful to show review questions in different fonts, like the [Jitai userscript](https://community.wanikani.com/t/jitai-%E5%AD%97%E4%BD%93-the-font-randomizer-that-fits/12617).

PR #63, opened 2018-10-22
Jitai fonts

hesch:
Hi again,

I implemented the random fonts requested in #30.

### Features:
- Choose random fonts to use from Settings
- Tap the japanese text in the review controller to switch back and forth between the random font and the standard font
- I added some of the fonts from the jitai page, but you can add more if you want. Just drag them into the Resources/fonts folder ;)

### Considerations:
- I updated the UserDefaults class to use NSCoding to store Objects (there is migration code present for the old NSStrings)
- I am not sure how to configure UITableView to use the height of the provided UITableViewCell. So I just set the height in the storyboard.
- I'm not sure about the licensing compatibility of this project and the fonts (the license files of the fonts are in japanese and I am still a noob)
- This time I didn't write tests because I have no Idea how "real" testing (mocking and stuff) in objective-c works. sorry :(

I hope you like the changes.


ios/ReviewViewController.m
```diff
@@ -24,6 +24,7 @@
 #import "TKMKanaInput.h"
 #import "proto/Wanikani+Convenience.h"
 #import "UIView+SafeAreaInsets.h"
+#import "TKMFontLoader.h"
```
davidsansome: Can you sort these imports alphabetically?

ios/TKMFontLoader.h
```diff
@@ -0,0 +1,29 @@
...
+@interface TKMFontLoader : NSObject
+
++ (NSArray*) getLoadedFonts;
++ (NSString*) getRandomFontToRender:(NSString*)text;
++ (void) saveToUserDefaults;
+
+@end
+
+@interface TKMFont : NSObject <NSCoding>
+  @property(strong) NSString *fontName;
```
davidsansome: strong is the default so I wouldn't bother putting it here.  You should put nonatomic though - the default, atomic, adds extra locking which is bad for performance if you don't need it.  Also unindent these two properties.

So it should be:

    @interface TKMFont ...
    
    @property(nonatomic) NSString *fontName;
    @property(nonatomic) BOOL enabled;
    
    @end

ios/TKMFontLoader.m
```diff
@@ -0,0 +1,125 @@
...
+
+#import "TKMFontLoader.h"
+
+#import <Foundation/Foundation.h>
+#import <CoreText/CoreText.h>
+
+#import "UserDefaults.h"
+
+static dispatch_once_t sOnceToken;
+static NSArray *loadedFonts;
```
davidsansome: Rename this sLoadedFonts so it's more obvious it's static

ios/TKMFontLoader.h
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import <CoreText/CoreText.h>
+
+@interface TKMFontLoader : NSObject
+
++ (NSArray*) getLoadedFonts;
```
davidsansome: Use lightweight generics wherever possible:

    + (NSArray<TKMFont *> *)getLoadedFonts;

You'll probably need to forward-declare the TKMFont class above, like:

    @class TKMFont;

ios/TKMFontLoader.h
```diff
@@ -0,0 +1,29 @@
...
++ (NSString*) getRandomFontToRender:(NSString*)text;
```
davidsansome: Just from looking at the API, it feels weird that this returns a string but the other one returns a font.  It's not clear what this string is.  How about renaming this method getRandomFontNameToRenderText:(NSString *)text?  Or changing it to return a TKMFont, then users would look at the .fontName property.

ios/TKMFontLoader.h
```diff
@@ -0,0 +1,29 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+#import <CoreText/CoreText.h>
+
+@interface TKMFontLoader : NSObject
```
davidsansome: This class doesn't have any instance methods or instance variables, so you could actually just make all its methods plain old C functions.  Objective-C classes add a bit of binary size and runtime overhead, so it's better to avoid them if you don't need them.

I would replace these with:

    NSArray<TKMFont *> *TKMGetLoadedFonts();
    NSString *TKMGetRandomFontToRender(NSString *text);
    void TKMSaveFontsToUserDefaults();

ios/TKMFontLoader.m
```diff
@@ -0,0 +1,125 @@
...
+
+void EnsureInitialized() {
+  dispatch_once(&sOnceToken, ^{
+    NSBundle *mainBundle = [NSBundle mainBundle];
+    NSArray *urls = [mainBundle URLsForResourcesWithExtension:nil subdirectory:@"Resources/fonts"];
```
davidsansome: NSArray<NSURL *> *

ios/TKMFontLoader.m
```diff
@@ -0,0 +1,125 @@
...
+    NSFileManager *fileManager = [NSFileManager defaultManager];
+    NSMutableArray *fonts = [[NSMutableArray alloc] init];
```
davidsansome: NSMutableArray has a convenience static constructor thing that I prefer to use:

    NSMutableArray<TKMFont *> *fonts = [NSMutableArray array];

ios/TKMFontLoader.m
```diff
@@ -0,0 +1,125 @@
...
+    
+    NSArray *fontsFromDefaults = UserDefaults.usedFonts;
```
davidsansome: NSArray<TKMFont *> *

ios/TKMFontLoader.m
```diff
@@ -0,0 +1,125 @@
...
+      NSData *data = [fileManager contentsAtPath: [url path]];
+      
+      CFErrorRef error;
+      CGDataProviderRef provider = CGDataProviderCreateWithCFData((CFDataRef)data);
+      CGFontRef font = CGFontCreateWithDataProvider(provider);
+      if (! CTFontManagerRegisterGraphicsFont(font, &error)) {
+        CFStringRef errorDescription = CFErrorCopyDescription(error);
+        NSLog(@"Failed to load font: %@", errorDescription);
+        CFRelease(errorDescription);
+      }
```
davidsansome: You'll need a continue in here too so you don't create a TKMFont for it if there was an error.
Also I think you need to CFRelease(provider) if there's an error as well.

How about moving this into static function:

    static NSString *FontNameFromURL(NSURL *url);

that returns nil if there was an error, then it'll be easier to see the logic in this method.

ios/TKMFontLoader.m
```diff
@@ -0,0 +1,125 @@
...
+      CFStringRef fontName = CGFontCopyFullName(font);
+      
+      TKMFont *newFont = [[TKMFont alloc] init];
+      newFont.fontName = (__bridge NSString *) fontName;
+      
+      NSPredicate *fontNamePredicate = [NSPredicate predicateWithFormat: @"fontName == %@", newFont.fontName];
+      NSArray *filteredArray = [fontsFromDefaults filteredArrayUsingPredicate:fontNamePredicate];
+      newFont.enabled = [filteredArray count] >= 1 ? [[filteredArray objectAtIndex:0] enabled] : false;
```
davidsansome: Thanks to objective-C's crazy behaviour with nils, you can do:

    newFont.enabled = filteredArray.firstObject.enabled;

If the array is empty, firstObject returns nil, and nil.enabled is NO.

ios/TKMFontLoader.m
```diff
@@ -0,0 +1,125 @@
...
+}
+
++ (NSString*)getRandomFontToRender:(NSString*)text {
+  NSPredicate *fontPredicate = [NSPredicate predicateWithBlock:^BOOL(TKMFont *font, NSDictionary *_) {
+    return font.enabled && [TKMFontLoader font:font canRender:text];
+  }];
+  
+  NSArray *availableFonts = [[TKMFontLoader getLoadedFonts] filteredArrayUsingPredicate:fontPredicate];
+  NSUInteger random = arc4random_uniform((uint32_t)[availableFonts count]);
+  return [[availableFonts objectAtIndex:random] fontName];
```
davidsansome: I would do availableFonts[random].fontName

ios/Tables/TKMFontModelItem.h
```diff
@@ -0,0 +1,38 @@
...
+
+@interface TKMFontModelItem : NSObject <TKMModelItem>
+
+- (instancetype)initWithFont:(TKMFont *)font NS_DESIGNATED_INITIALIZER;
+- (instancetype)init NS_UNAVAILABLE;
+
+- (void)setSelected:(BOOL)selected;
+
+@property(nonatomic, copy) TKMFont *font;
+@property(nonatomic, weak) id delegate;
```
davidsansome: I don't think you use this delegate?

ios/Tables/TKMFontModelItem.m
```diff
@@ -0,0 +1,66 @@
...
+- (void)updateWithItem:(TKMFontModelItem *)item {
+  [super updateWithItem:item];
+  
+  _fontNameLabel.text = item.font.fontName;
+  NSUInteger oldSize = _fontPreviewLabel.font.pointSize;
+  _fontPreviewLabel.font = [UIFont fontWithName:item.font.fontName size:oldSize];
+  _fontPreviewLabel.text = @"あいうえお\n漢字 字体";
+}
+
+- (void)awakeFromNib {
```
davidsansome: Delete this method :)

ios/UserDefaults.h
```diff
@@ -43,6 +43,8 @@ DECLARE_BOOL(animatePlusOne);
 
 // Review settings.
 DECLARE_ENUM(ReviewOrder, reviewOrder);
+DECLARE_BOOL(randomFontsEnabled);
+DECLARE_OBJECT(NSArray, usedFonts);
```
davidsansome: NSArray<TKMFont *> might work here too!

ios/TKMFontLoader.h
hesch: Do you think I should put these Functions in a separate file and rename TKMFontsLoader to TKMFont?

ios/ReviewViewController.m
hesch: How do you want them sorted? Should `#import "proto/Wanikani+Convenience.h"` be sorted by filename or path?

hesch:
You are right about the fonts. I took them from the original [Jitai plugin](https://gist.github.com/obskyr/9f3c77cf6bf663792c6e#fonts-from-freejapanesefontcom). I just unpacked the archives and removed any unnecessary file. Some of the websites were down at the time I looked at them, but now they seem to be up again. Should we add the other fonts too?

What do you think about the font preview? Should I add other/more characters or make the preview area and/or font size bigger? Unfortunately I'm not in the paid apple developer program, so I can't get a real feel how the preview looks an a real device.

davidsansome:
> Should we add the other fonts too?

Yes I think so!  Although some of them are pretty big, I'd worry about increasing the binary size too much.  Maybe later on we can think about downloading them on demand only for the users that want them.

> What do you think about the font preview? Should I add other/more characters or make the preview area and/or font size bigger?

I think it's okay for now.  I'd like to make it more beautiful later, but I can't think of any good ideas at the moment :)

> Unfortunately I'm not in the paid apple developer program, so I can't get a real feel how the preview looks an a real device.

You can still install apps on your device without paying!  You just can't upload them to the App Store, and any app you install disappears after 7 days.  Just sign in to Xcode with your (free) Apple ID.  You might also have to change the bundle ID in the project settings - not sure.

davidsansome:
choose-fonts

How about this for the font chooser view?  All the fonts would start out "not installed" (3rd row).  Tapping on one would start downloading it (4th row).  When it finishes downloading it is selected (1st row).  Tapping on it de-selects it (2nd row).  Tapping a de-selected one selects it again.  Long-pressing an installed font pops up a menu with an option to delete it.

We'd have to figure out how to render the font previews without the fonts being installed.  We could probably pre-render every font on every screen size using the simulator, and bundle the .pngs with the app.

davidsansome:
I've just done a similar downloading UI for audio: https://github.com/davidsansome/tsurukame/blob/master/ios/TKMOfflineAudioViewController.mm.  I'll try to make that more generic so we can have some common code that's shared between audio downloads and font downloads.  As for hosting, I've set up https://tsurukame.app that's hosted on Firebase - configs are here: https://github.com/davidsansome/tsurukame/tree/master/www.

> Shouldn't the delete action be on a swipe gesture or on an edit icon in the top right corner?

Yeah I think the swipe gesture/edit icon is a better idea!

I've merged your commits so we won't get any more merge conflicts.  I'm happy to have a go at doing the download stuff this week!

PR #84, opened 2018-12-11
Review notifications

### LINKED ISSUE #64: Local Push Notifications
Would be really useful to have local notifications sent when there are new reviews, to coincide with the badge update.

indirect:
Right now, the only thing keeping Allicrab installed on my phone is the notifications when new reviews or lessons are available. This PR adds local notifications at the same time as the badge number updates to show how many reviews are available.

Closes #64.

### DIFF
--- a/ios/AppDelegate.m
+++ b/ios/AppDelegate.m
@@ -150,7 +150,7 @@
         nextDateAfterDate:[NSDate date]
              matchingUnit:NSCalendarUnitMinute
                     value:0
-                  options:NSCalendarMatchNextTime | NSCalendarMatchStrictly];
+                  options:NSCalendarMatchNextTime];
     NSTimeInterval startInterval = [startDate timeIntervalSinceNow];
     int cumulativeReviews = reviewCount;
     for (int hour = 0; hour < upcomingReviews.count; hour++) {
--- a/ios/AppDelegate.m
+++ b/ios/AppDelegate.m
@@ -167,6 +167,7 @@
       }
       NSString *identifier = [NSString stringWithFormat:@"badge-%d", hour];
       UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
+      content.body = [NSString stringWithFormat:@"%d review%@ available", cumulativeReviews, cumulativeReviews == 1 ? @"" : @"s"];
       content.badge = @(cumulativeReviews);
       UNNotificationTrigger *trigger =
           [UNTimeIntervalNotificationTrigger triggerWithTimeInterval:triggerTimeInterval


ios/AppDelegate.m
```diff
@@ -167,6 +167,10 @@ - (void)updateAppBadgeCount {
       }
       NSString *identifier = [NSString stringWithFormat:@"badge-%d", hour];
       UNMutableNotificationContent *content = [[UNMutableNotificationContent alloc] init];
+      content.body = [NSString stringWithFormat:@"%d reviews available", cumulativeReviews];
+      if (cumulativeReviews == 1) {
+        content.body = [content.body stringByReplacingOccurrencesOfString:@"reviews" withString:@"review"];
```
davidsansome: I'd prefer to do this all in one go, rather than building the string then modifying it.

    content.body = [NSString stringWithFormat:@"%d review%@ available", cumulativeReviews, cumulativeReviews == 1 ?: @"s"];

ios/AppDelegate.m
```diff
@@ -153,7 +153,7 @@ - (void)updateAppBadgeCount {
                   options:NSCalendarMatchNextTime | NSCalendarMatchStrictly];
     NSTimeInterval startInterval = [startDate timeIntervalSinceNow];
     int cumulativeReviews = reviewCount;
-    for (int hour = 0; hour < upcomingReviews.count; hour++) {
+    for (int hour = 1; hour < upcomingReviews.count; hour++) {
```
davidsansome: I thought hour 0 was the next hour boundary.  So if it's 10:25 now, hour 0 would represent 11:00.  I think you'd still need to include this.

ios/AppDelegate.m
```diff
@@ -167,6 +167,10 @@ - (void)updateAppBadgeCount {
       }
       NSString *identifier = [NSString stringWithFormat:@"badge-%d", hour];
```
davidsansome: Since each notification has a unique identifier, will this potentially add a new notification for every hour that there are new reviews?  So after 12 hours I'll have 12 notifications in my tray?
Is there any to have these replace the contents of the existing notification when they're triggered?  I'm not sure what'll happen if you add them all with the same identifier - the docs make it sound like it'll only keep one pending notification per identifier.  What about threadIdentifier?

ios/AppDelegate.m
indirect: Seems reasonable, I'll update it.

ios/AppDelegate.m
indirect: Ahhh, okay. If hour 0 is the upcoming hour boundary, then we definitely need to adjust the time that the notification will be posted at. I'll do that too.

ios/AppDelegate.m
indirect: iOS 12 will automatically group all of the notifications together, but if you want to always replace previous notifications we could use the same identifier for all of them, which the docs say will result in one notification in the notification center.

indirect: After some additional research, it looks like you are right... the only way to remove already-delivered local notifications is [removeAllDeliveredNotifications](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649502-removealldeliverednotifications) or removing a single notification by identifier. I'm personally completely ok with multiple notifications appearing in a stacked group in the notification center, but if that's not okay maybe we can send only one notification at the first time new reviews are available?

davidsansome: Thanks for checking!  I think multiple notifications will be okay, particularly since iOS 12 groups them all together.
But it might be nice to have an option to notify the user once per day when they get over some threshold: for example "notify me when I have > 50 reviews pending", or "notify every day at 8am".  That'll suit people who are more laid back and aren't bothered about doing all the reviews as soon as they're available (like me).

PR #88, opened 2019-01-23
Add a setting to hide old mnemonics from radicals.

cpojer:
I recently started using WaniKani and Tsurukame. I noticed that there have been changes to the mnemonics for radicals over the years and old mnemonics are still shown in the app, possibly for existing users. While learning new radicals as a new user, I can't help but also look at the old mnemonics and I tend to get confused.

In this PR I am adding a new setting "Show Old Mnemonics" that is enabled by default which keeps the current behavior of the app. When disabled it will hide the old meanings both from the levels view as well as the detail view. I did not make them invalid as answers because it is only meant to be a visual change. It will also ensure that future updates to radical mnemonics won't disrupt current users.

## Test Plan
Click this image for a [video](https://youtu.be/soIZ85Mu-CQ) showing this feature:
[test video](https://youtu.be/soIZ85Mu-CQ)

Finally, I just wanna thank you for building such a high quality app, making it free and open source to contribute. Tsurukame is one of my favorite iOS apps. Thank you!

### DIFF
--- a/ios/SettingsViewController.m
+++ b/ios/SettingsViewController.m
@@ -110,6 +110,12 @@ typedef void (^NotificationPermissionHandler)(BOOL granted);
                                                         on:UserDefaults.enableCheats
                                                     target:self
                                                     action:@selector(enableCheatsSwitchChanged:)]];
+  [model addItem:[[TKMSwitchModelItem alloc] initWithStyle:UITableViewCellStyleSubtitle
+                                                     title:@"Show old mnemonics"
+                                                  subtitle:@"Display old mnemonics alongside new ones"
+                                                        on:UserDefaults.showOldMnemonic
+                                                    target:self
+                                                    action:@selector(showOldMnemonicChanged:)]];
 
   [model addSection:@"Audio"];
   [model addItem:[[TKMSwitchModelItem alloc]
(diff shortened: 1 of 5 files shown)


ios/SubjectDetailsView.m
```diff
@@ -68,9 +69,10 @@
     };
     [strings addObject:[[NSAttributedString alloc] initWithString:meaning attributes:attributes]];
   }
-  for (TKMMeaning *meaning in meanings) {
+  for (TKMMeaning *meaning in subject.meaningsArray) {
     if (meaning.type != TKMMeaning_Type_Primary &&
-        meaning.type != TKMMeaning_Type_Blacklist) {
+        meaning.type != TKMMeaning_Type_Blacklist &&
+        (meaning.type != TKMMeaning_Type_AuxiliaryWhitelist || !subject.hasRadical || UserDefaults.showOldMnemonic)) {
```
cpojer: While `TKMMeaning_Type_AuxiliaryWhitelist` should only ever apply to radicals, I chose to explicitly check for `subject.hasRadical` for now to guard for potentially unexpected API changes to kanji (or vocabulary).

ios/SettingsViewController.m
```diff
@@ -110,6 +110,12 @@ - (void)rerender {
                                                         on:UserDefaults.enableCheats
                                                     target:self
                                                     action:@selector(enableCheatsSwitchChanged:)]];
+  [model addItem:[[TKMSwitchModelItem alloc] initWithStyle:UITableViewCellStyleSubtitle
+                                                     title:@"Show old mnemonics"
+                                                  subtitle:@"Display old mnemonics alongside new ones"
```
cpojer: Happy to change the text here if you'd prefer something else.

davidsansome: That text sounds good to me!

Issue #31, opened 2018-05-02
Use iOS 'Adaptive Layout' to make app look better on larger screens for both iPhone and iPads.


mztik:
Kanji font size can be quite small. An option to increase the font size during lessons and reviews would be really welcomed by many users. As an example of this, please take a look at the attached screenshots of the app running on an iPad.

![screen shot 2018-05-02 at 7 44 55 pm](https://user-images.githubusercontent.com/169478/39554541-6ac9afd4-4e41-11e8-872c-cbe88bba895f.png)

davidsansome:
I really didn't think about tablets at all while designing this app.  It needs to work better on iPads and larger screens by default - I don't think a font size setting is the right fix.

mztik:
@davidsansome Please give this more serious thought. Read more on iOS "Adaptive Layout".

zoglesby:
![e11fe5e2-8cd5-43f0-9c06-b6706d50ff2f](https://user-images.githubusercontent.com/33433/43651514-a4d990a0-9710-11e8-8824-2a59f552199c.png)

I am sure it’s know but this is what the app currently looks like on the iPad. The header is too small for the screen.

turtlemaster19:
![Review](https://user-images.githubusercontent.com/46784000/67484177-24a0f400-f62d-11e9-9fac-d82b8e8f8013.jpeg)
![Lesson](https://user-images.githubusercontent.com/46784000/67484178-24a0f400-f62d-11e9-802b-779569e7be4e.jpeg)
@davidsansome, this issue is resolved. I can't find the pull request that fixed it, but these images look much better than the above in this thread.

turtlemaster19:
See #37. Regardless, this issue is clearly closed.

PR #102, opened 2019-03-14
Boost iPad question font size

indirect:
Partially addresses #31 by making the question font size bigger on iPads.


davidsansome:
Thanks for fixing this!

PR #103, opened 2019-03-14
Advance on "return" after wrong answer

### LINKED ISSUE #95: Enter on the iPad Pro keyboard doesn’t work after an invalid answer
I see a vocabulary kanji. I put in “sightseeing” when the answer should be “Idea”. the app tells me it’s wrong. I press enter on my iPad Pro 2016 12.9” smart keyboard, but it no longer advances to the next review. On version 1.9.

indirect:
This PR resolves #95 by allowing ReviewViewController to become the first responder itself. That way, after the `answerField` resigns first responder on line 644, the ReviewViewController can still respond to the return key and move on to the next question.

https://github.com/indirect/tsurukame/blob/74b9c80ab058a45bd0e6aa02ef6434ef810f0d60/ios/ReviewViewController.mm#L644

### DIFF
--- a/ios/ReviewViewController.mm
+++ b/ios/ReviewViewController.mm
@@ -280,6 +280,10 @@ class AnimationContext {
   return (int)_activeQueue.count;
 }
 
+- (BOOL)canBecomeFirstResponder {
+  return true;
+}
+
 #pragma mark - UIViewController
 
 - (void)viewDidLoad {


indirect:
(not super related, but the reason I can't base these PRs on the tip of master is because there are files missing and the build fails. @davidsansome, could you do a git add and push all the files needed to build the tip of `master`?)

PR #104, opened 2019-03-14
Enlarge font in reviews also

indirect:
For reasons I don't really understand, increasing the question label
font size inside `viewDidLoad` only makes the question appear bigger
during lesson quizzes, and does not make the question appear bigger
during reviews. I think it might be because of the setupContext block,
but I wasn't able to fix it by changing that.

Setting the font size to the correct size in `viewWillLayoutSubviews`
always seems to work, so I'm now doing that.

This patch increases the font size for both custom fonts and the system
font, in both lessons and reviews, both when tapping on the question
label and when not tapping on the question label.

### DIFF
--- a/ios/Main.storyboard
+++ b/ios/Main.storyboard
@@ -609,7 +609,7 @@
                                 <nil key="highlightedColor"/>
                             </label>
                             <button opaque="NO" contentMode="scaleAspectFit" contentHorizontalAlignment="left" contentVerticalAlignment="center" buttonType="roundedRect" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="ctF-hm-AxT">
-                                <rect key="frame" x="307" y="601.5" width="68" height="67"/>
+                                <rect key="frame" x="278" y="588" width="97" height="94"/>
                                 <accessibility key="accessibilityConfiguration" label="Confirm"/>
                                 <inset key="contentEdgeInsets" minX="20" minY="20" maxX="20" maxY="20"/>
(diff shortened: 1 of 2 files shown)


davidsansome:
Sorry for the delay in replying to this one!  I've been without my mac this week and I really wanted to test it out to see what's going on.

I think your problem was actually that traitCollection.horizontalSizeClass was returning UIUserInterfaceSizeClassUnspecified on the very first randomTask that gets called from viewDidLoad - before anything is added to the view hierarchy and before it has a traitCollection set.  So it was returning the normal size, not the *2.5 size.
I found some details in this stackoverflow answer: https://stackoverflow.com/questions/30007701/view-traitcollection-horizontalsizeclass-returning-undefined-0-in-viewdidload/30009266

Doing it in viewWillLayoutSubviews messes with the question label animation a bit.  Try it with slow animations in the simulator - you'll notice the new label fade in immediately at its final position, without animating there from its old position.

I'd suggest using UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad instead to check whether you're on an iPad, and removing the viewWillLayoutSubviews override.

davidsansome:
This was what I had in mind: https://github.com/davidsansome/tsurukame/commit/d307cfd307ecf299d860cf6aae8855812e3ecff0
If that sounds OK to you I can just merge your PR and commit that on top!

indirect:
@davidsansome ahh, brilliant! Go for it. :)

indirect:
Rebased and added your fix on top, this should be good to merge now! 👍

PR #109, opened 2019-03-28
Add ability to customize lessons

cpojer:
When leveling up, a lot of radicals, kanji and vocabulary become available at the same time. I usually tend to study ~5 new kanji per day which means that it often happens that I'll only get to study vocabulary once I am pretty close to leveling up again. I prefer to work myself through both kanji and vocabulary at the same time to reduce the load of new words towards the end of a level. Currently I use the wanikani website with a custom userscript but I'd prefer to use Tsurukame exclusively because of the much better user experience.

This PR adds two new options to customize which items are taught first in the lessons screen:
* Prioritize current level (default: disabled): When enabling this option, items from the current (highest) level will be shown before any lower levels.
* Sort by subject type (default: Radical, Kanji, Vocabulary): This option navigates to a new screen.

## Screenshot of the Settings view:
Screen Shot 2019-03-27 at 23 55 48

## Screenshot of the subject type sorting view:
Screen Shot 2019-03-27 at 23 56 15

*Note: I am not an Objective-C expert and most of the code I wrote here is the first time I did certain things, like creating a sortable TableView. Please let me know how I could do things better!*

### DIFF
--- a/ios/LessonOrderViewController.m
+++ b/ios/LessonOrderViewController.m
@@ -1,0 +1,61 @@
+// Copyright 2018 David Sansome
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import "LessonOrderViewController.h"
+#import "UserDefaults.h"
+
+@interface LessonOrderViewController ()
+@end
+
+@implementation LessonOrderViewController
+
+- (void)viewDidLoad {
+  [super viewDidLoad];
+  if (![UserDefaults.lessonOrder count]) {
+    UserDefaults.lessonOrder = [[NSMutableArray alloc] initWithObjects:@"Radicals",@"Kanji",@"Vocabulary",nil];
+  }
+
+  self.tableView.editing = YES;
+}
+
+- (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section {
+  return UserDefaults.lessonOrder.count;
+}
+
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath {
+  UITableViewCell *cell = [self.tableView dequeueReusableCellWithIdentifier:@"cellIdentifier"];
+  if(cell == nil) {
+    cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleDefault reuseIdentifier:@"cellIdentifier"];  
+  }
+  cell.textLabel.text = [UserDefaults.lessonOrder objectAtIndex:indexPath.row];
+  return cell;
+}
+
+- (UITableViewCellEditingStyle)tableView:(UITableView *)tableView editingStyleForRowAtIndexPath:(NSIndexPath *)indexPath {
+  return UITableViewCellEditingStyleNone;
+}
+
+- (BOOL)tableView:(UITableView *)tableView shouldIndentWhileEditingRowAtIndexPath:(NSIndexPath *)indexPath {
+  return NO;
+}
+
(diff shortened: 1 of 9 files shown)


ios/LessonOrderViewController.m
```diff
@@ -0,0 +1,61 @@
...
+- (BOOL)tableView:(UITableView *)tableView shouldIndentWhileEditingRowAtIndexPath:(NSIndexPath *)indexPath {
+  return NO;
+}
+
+- (void)tableView:(UITableView *)tableView moveRowAtIndexPath:(NSIndexPath *)sourceIndexPath toIndexPath:(NSIndexPath *)destinationIndexPath {
+  NSMutableArray *copy = [[NSMutableArray alloc] initWithArray:UserDefaults.lessonOrder copyItems:YES];
+  NSString *item = [copy objectAtIndex:sourceIndexPath.row];
+  [copy removeObject:item];
+  [copy insertObject:item atIndex:destinationIndexPath.row];
+  UserDefaults.lessonOrder = copy;
```
cpojer: *I have no idea what I'm doing!*

Because of the ways the UserDefaults macros work, `UserDefaults.lessonOrder` can only be changed by assigning a new value, so operations like `insertObject` won't work. I realize copying the array, modifying it, and then writing it back is probably a bit crazy but if there is a better way to do this I'm happy to change it.

ios/ReviewItem.m
```diff
@@ -48,6 +49,17 @@ @implementation ReviewItem
   return ret;
 }
 
+- (NSUInteger)getSubjectTypeIndex:(TKMSubject_Type)type {
+  if (type == TKMSubject_Type_Radical) {
+    return [UserDefaults.lessonOrder indexOfObject:@"Radical"];
+  } else if (type == TKMSubject_Type_Kanji) {
+    return [UserDefaults.lessonOrder indexOfObject:@"Kanji"];
+  } else if (type == TKMSubject_Type_Vocabulary) {
+    return [UserDefaults.lessonOrder indexOfObject:@"Vocabulary"];
```
cpojer: Admittedly this is looking up the index for string values for every comparison, which is not a good way of doing it. I could change the `UserDefaults.lessonOrder` to contain the enum id but I couldn't figure out how to make that work well with the TableView. In practice we are only doing very few comparisons (<50 on average) so it should be fine but I'm happy to change that with some pointers.

ios/ReviewItem.m
```diff
@@ -59,20 +71,35 @@ - (instancetype)initFromAssignment:(TKMAssignment *)assignment {
 }
 
 - (NSComparisonResult)compareForLessons:(ReviewItem *)other {
-#define COMPARE(field)            \
```
cpojer: I didn't think this provided much value and would have become a ton more confusing so I instead just spelled it all out in the new version.

indirect:
This is amazing and I was thinking about trying to do this myself, so thank you! 🙌

indirect:
I've been using this for the last week, and it's been working great for me. 👍

cpojer:
Thanks for testing it out. I've also been using it for a week and it works pretty smoothly.

cpojer:
Wanted to highlight the value that this provides to me. After three days on the new level I was able to go through all 33 new vocabulary lessons while making good progress on kanji. Without this change, once I unlock the remaining kanji tomorrow, I would not have a chance to study vocabulary up until I'm done learning all the kanji for the level. With the change proposed in this PR I can pace myself much better and study ~10 pieces of vocabulary alongside ~5 kanji every day.

IMG_4766

ios/LessonOrderViewController.m
davidsansome: It's the property setter that does the saving, so I think you could get away with getting the array, insertObject on it, and then setting it back again.

    [UserDefaults.lessonOrder insertObject:...];
    UserDefaults.lessonOrder = UserDefaults.lessonOrder;

But that's a bit nasty as well.

I'd actually prefer the one in UserDefaults to be non-mutable, and then keep doing this copy, modify, assign here.

ios/ReviewItem.m
davidsansome: I think it would be cleaner to store the enum values in `UserDefaults.lessonOrder`, and then turn them back into strings in `tableView:cellForRowAtIndexPath:`.
You'd need to have a switch like:

    switch ([UserDefaults.lessonOrder objectAtIndex:indexPath.row]) {
      case TKMSubject_Type_Radical:
        cell.textLabel.text = @"Radicals";
        break;
      ...

You'll probably also need to box the values in NSNumber to put them in the NSArray.

PR #111, opened 2019-03-30
Add space as keyboard shortcut to play reading

indirect:
This integrates well with keyboard control for navigating lessons, making it possible to learn new subjects (including hearing their readings) without having to tap on the screen.


indirect:
Oops. Don't merge this, yet! I just discovered that it stops you from being able to type a space when you are answering the lesson quiz. 😅

indirect:
Okay, safe to merge! Lesson quiz no longer eats spaces. 😆

PR #114, opened 2019-04-06
Add keyboard shortcuts for cheating

indirect:
This builds on top of #106, adding keyboard shortcuts not just to play the reading audio, but also keyboard shortcuts to *A*sk again, Mark *C*orrect, and Add *S*ynonym. The shortcuts show up when you hold down the command key on the "missed an answer" screen:

IMG_1A0F88007F51-1


davidsansome:
Thanks for all these!  Sorry it took me so long to look at them!

PR #121, opened 2019-05-13
Persist SubjectCatalogueVC's show answers toggle

maddievision:
This creates a user default that isn't actually used in the settings view. Instead it is controlled and read directly by the show answers toggle in the SubjectCatalogueViewController. I found myself not really wanting to see the answers and was wanting a way to have it be off when I enter the view!

I'm not sure if this breaks any kind of paradigm since I don't think there are any existing settings that persist outside of the settings view, so this might throw someone off. If it is preferred to be maintained from there, I can move it there, where it could be in a new section called "Catalogue" and named "Show answers by default" or something.

Let me know what you think!

### DIFF
--- a/ios/SubjectCatalogueViewController.m
+++ b/ios/SubjectCatalogueViewController.m
@@ -16,6 +16,7 @@
 #import "DataLoader.h"
 #import "SubjectsByLevelViewController.h"
 #import "TKMServices.h"
+#import "UserDefaults.h"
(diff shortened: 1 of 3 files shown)


davidsansome:
Great idea!  I think it's fine to have this in UserDefaults - it's for all kinds of settings, not only the ones that can be set in the settings view.

Issue #135, opened 2019-07-08
Crash on iOS 13 due to missing Hiragino Sans weights


indirect:
I'm not sure if this is iOS 13-specific, but Tsurukame crashes on my iPad when this line returns nil: 

https://github.com/davidsansome/tsurukame/blob/d813ea063f40439388aff897f9b8f4835a4652f3/ios/Style.m#L106

According to the debugger, the only weights available are W3, W6, and W8:

```
(lldb) po [UIFont fontNamesForFamilyName:@"Hiragino Sans"]
<__NSArrayM 0x2802ef420>(
HiraginoSans-W3,
HiraginoSans-W6,
HiraginoSans-W8
)
```

I guess I'll send a PR to fall back on W3 and W8 if W2 and W7 aren't there? 😬

indirect:
Huh, according to Apple's [system fonts page](https://developer.apple.com/fonts/system-fonts/), iOS has _never_ had any weights except 3, 6, and 8:

<img width="1057" alt="image" src="https://user-images.githubusercontent.com/78/60776173-e9f76800-a0de-11e9-8ab4-79c0d3bb0f3d.png">

PR #136, opened 2019-07-08
Use font weights that iOS has

indirect:
This is a fix for #135, and switches to the font weights that are included with iOS according to both Apple's System Fonts page and real-life listing of available fonts as detailed in that issue's comments.


davidsansome:
Thanks!  I'm not sure why this worked for me on iPhone!

Issue #122, opened 2019-05-14
Reviews aren't being uploaded even though I'm online. 


komocode:
![IMG_0692](https://user-images.githubusercontent.com/226526/57737453-9decf180-7660-11e9-94c0-afc5babf71da.png)

davidsansome:
Thanks for reporting this!  It's a problem caused by a recent change to WaniKani that started expiring auth cookies.

I think I've fixed it now!  Would you mind installing a new version from testflight, and let me know if it works?  https://testflight.apple.com/join/Fijye2AA

komocode:
I installed it just now without deleting the app. It seems to still show "225 review progress:

indirect:
Sadly I am also seeing this, running from the latest commit ca4cad025e848ad2947411e14b3e89764c8808f3.

indirect:
Hmm... I have no idea, now. I installed a build without the API changing commit, and it was able to upload pending reviews. Then I deleted the app, re-installed the exact previous build that couldn't upload pending reviews, and... it uploaded a new set of pending reviews just fine. 😬

lyallcooper:
I experienced the same issue with the testflight build, where it couldn't upload lesson/review progress, but reverting to the app store version uploaded them fine. However after reinstalling the testflight build it _still_ couldn't upload progress, unlike @indirect.

indirect:
To be clear, it seems to work for me if I delete the app entirely before installing the new build with this change—upgrading in place by installing this change over the old build seems to make things get stuck. At least for me.

indirect:
My iPad got into a super weird state: 4 reviews queued to be uploaded "when I'm back online", but doing new reviews immediately uploaded them. Without uploading those 4 queued reviews. I'm very sure the "live" reviews were making it to the server because I got an level-up email from Koichi while I was testing to see if new reviews would get uploaded... but those 4 queued reviews never did get uploaded. I had to log out and log back in to make them go away.

indirect:
Oh weird... logging out and then logging back in seems to have gotten it into a state where nothing gets uploaded to the server and everything gets queued. 😲

indirect:
As far as I can tell, this was fixed for me by #138. Is anyone else still seeing it?

brookshelley:
Just started seeing this with the latest update—I’m online and 82 reviews are showing as “will be uploaded when you’re back online”

indirect:
[...]

PR #138, opened 2019-07-12
Discard explicitly rejected progress

indirect:
I think this is the solution for eternally queued updates I was seeing, that I mentioned in #122. 

It turns out it's possible to start reviews before the app has caught up with reviews that you did on another device, especially if the other device didn't finish reporting reviews before it suspended. In that case, one of the devices ends up with progress waiting to report that can't ever be accepted, because the server rejects the request with 422 Unprocessable Entity.

Since the server is telling is that there is no point in the future where this particular progress report will _ever_ be accepted, I'm using that as my queue to simply delete that progress from the database and call it good. Plus, installing this commit fixed all my stuck queued progress. :)


davidsansome:
Wow nice find!  Thanks!

PR #139, opened 2019-07-13
SRS level counts on dashboard

indirect:
I use the SRS level counts (along with reviews coming up in the next 24 hours) to know whether I feel ok about doing more lessons. Since it's super helpful to me, I thought it might be helpful to other people as well!

Screen Shot 2019-07-13 at 3 29 08 PM


ios/LocalCachingClient.mm
```diff
@@ -570,34 +571,69 @@ - (void)maybeUpdateAvailableSubjectCounts {
...
+  if (!_isCachedSrsLevelCountsStale) {
     return;
   }
 
-  _cachedGuruSubjectCounts.fill(0);
+  _cachedSrsLevelCounts.fill(0);
   [_db inDatabase:^(FMDatabase *_Nonnull db) {
+    FMResultSet *gr = [db executeQuery:
+                             @"SELECT COUNT(*) FROM subject_progress WHERE srs_stage "
+                             @">= 5 AND subject_type = ? GROUP BY subject_type",
```
davidsansome: You don't need the GROUP BY subject_type because you're filtering it to only Kanji

ios/LocalCachingClient.mm
```diff
@@ -570,34 +571,69 @@ - (void)maybeUpdateAvailableSubjectCounts {
...
+                             @"SELECT srs_stage, COUNT(*) FROM subject_progress "
+                             @"WHERE srs_stage >= 2 GROUP BY srs_stage"];
     while ([r next]) {
-      int subject_type = [r intForColumnIndex:0];
+      int srs_stage = [r intForColumnIndex:0];
       int count = [r intForColumnIndex:1];
-      if (subject_type >= _cachedGuruSubjectCounts.size()) {
-        continue;
+      int srs_level = 0;
+      switch(srs_stage) {
```
davidsansome: Could you move this logic to proto/Wanikani+Convenience.m?  There are already functions there that do something similar, it would be good to keep it all together.
Also it would be nice to give these srs_level values names in an enum.

I'm thinking something like:

```
enum TKMSRSStageCategory {
  TKMSRSStageNovice = 0,
  TKMSRSStageApprentice = 1,
  TKMSRSStageGuru = 2,
  ...
}

TKMSRSStageCategory TKMSRSStageCategoryForLevel(int srs_stage) {
  switch (srs_stage) {
    case 2:
    case 3:
    case 4:
      return TKMSRSStageApprentice;
    ...
}
```

Then you can use that TKMSRSStageCategory type in the function signature (- (int)getSrsLevelCount:(TKMSRSStageCategory)level)

ios/LocalCachingClient.mm
indirect: oh good point 😁

ios/LocalCachingClient.mm
indirect: I just tried doing that, and it seems like it works, except... the linker says the symbol for the new function is missing? I have no idea. I tried pushing it, though. 😬

indirect: I did some more testing and trying to call any of the functions from the convenience file in the localCachingClient breaks the linker. The only difference I can find is that localCachingClient is a .mm file instead of a .m file? But I have no idea why adding C++ classes would make a symbol seem missing. @davidsansome any ideas here?

indirect:
Pushed my commit with the linker error, and the linker error shows up on my other machine, so I think you'll probably be able to reproduce it? Would love suggestions about how to either fix the linker error or just move the code somewhere else that the linker won't complain about.

ios/LocalCachingClient.mm
davidsansome: Oh yeah, that's exactly the problem - you need to surround everything in the .h with:

```
#ifdef __cplusplus
extern "C" {
#endif

...

#ifdef __cplusplus
}
#endif
```

See https://github.com/davidsansome/tsurukame/blob/master/ios/Style.h#L19 for an example and https://en.wikipedia.org/wiki/Name_mangling#Handling_of_C_symbols_when_linking_from_C++ for an explanation why :)

indirect: Ohhhh yes that is clearly the issue, thanks! I’ll fix it shortly.

I’m working on this project partly as a way to ease in to iOS development, so I’d be happy to help convert to Swift if that’s something you’re interested in. I’m... marginally competent at both Obj-C and Swift, but have spent more hours and shipped more code written in Swift at this point.

indirect: Phew! Got it working, after debugging that I was actually defining a function with a different name than the symbol I was declaring and trying to use, so the linker was actually right. 😂

indirect:
ok! all set, and rebased against the latest master. 👍

davidsansome:
Thanks again for this, it looks really nice!
I'm thinking it's about time to make another release - have you got any more changes in the works for me or should I cut a release here?

indirect:
The "Time remaining" cell in the screenshot above didn't make this PR because it wasn't quite working. I have it working now, but it's not possible to calculate a level-up date if there are any locked items on the level. I'm noodling with fetching `level_progressions` to calculate the average level up time and show that if we can't show a definitive time yet, but it's been harder than I was hoping it would be.

Sounds like you can either:
- Release now, which would at least get the latest dataset out there
- Merge the "Time remaining" cell without average times (just showing "Locked" if items are still locked)
- Wait for me to finish the average time remaining bit

I'm ok with any of those, because I'm running my own build anyway ¯\_(ツ)_/¯

davidsansome:
Sure let's wait for the average time remaining bit to be done - there's no hurry.
Thanks again for your work 😄

PR #140, opened 2019-07-14
Add "Show remaining" screen showing level progress

indirect:
This is the screen I've always wanted in Tsurukame, and I think means that I won't need to keep AlliCrab installed anymore (😅). The "Show remaining" screen shows the radicals and kanji that have not yet reached guru status on the current level, along with the next times each item will be reviewable and could potentially reach guru. Here's what it looks like:

Simulator Screen Shot - iPhone X - 2019-07-14 at 14 48 03

Simulator Screen Shot - iPhone X - 2019-07-14 at 01 27 50

Simulator Screen Shot - iPhone X - 2019-07-14 at 01 28 04

### DIFF
--- a/ios/Tables/TKMSubjectModelItem.m
+++ b/ios/Tables/TKMSubjectModelItem.m
@@ -103,16 +103,17 @@ static const CGFloat kFontSize = 14.f;
 
  switch (itemLevel) {
     case 1:
-      guruInterval += (isAccelerated ? 7200 : 14400);
+      guruInterval += (isAccelerated ? 2 : 4);
     case 2:
-      guruInterval += (isAccelerated ? 14400 : 28800);
+      guruInterval += (isAccelerated ? 4 : 8);
     case 3:
-      guruInterval += (isAccelerated ? 28800 : 82800);
+      guruInterval += (isAccelerated ? 8 : 23);
     case 4:
-      guruInterval += (isAccelerated ? 82800 : 169200);
+      guruInterval += (isAccelerated ? 23 : 47);
   }
 
-  return [reviewDate dateByAddingTimeInterval:guruInterval];
+  int guruSeconds = guruInterval * 60 * 60;
+  return [reviewDate dateByAddingTimeInterval:guruSeconds];
 }
 
 @end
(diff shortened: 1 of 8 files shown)


indirect:
I'm... very unsure... about whether the structure of this code is any good at all, but it's working for me, so I figured I'd post it. I am very happy to make changes if requested.

ios/Tables/TKMSubjectModelItem.m
```diff
@@ -65,6 +66,55 @@ - (NSString *)cellNibName {
...
+  if (itemLevel == guruLevel) {
+    return reviewDate;
+  }
+
+  double guruInterval = 0;
+  bool isAccelerated = _subject.level <= 2;
+
+ switch (itemLevel) {
+    case 1:
+      guruInterval += (isAccelerated ? 7200 : 14400);
```
davidsansome: I think these numbers would be easier to understand if they were in hours * 60 * 60, like:

```
guruInterval += (isAccelerated ? 2 * 60 * 60 : 4 * 60 * 60);
```

Or you could specify them all in hours and do the conversion to seconds at the bottom:

```
double guruIntervalHours = 0;
...
guruIntervalHours += (isAccelerated ? 2 : 4);
...
return [reviewDate dateByAddingTimeInterval:guruIntervalHours * 60 * 60];
```

Also probably a good idea to put a link to the docs in a comment:

```
// From https://docs.api.wanikani.com/20170710/#additional-information
```

ios/Tables/TKMSubjectModelItem.m
```diff
@@ -99,25 +149,68 @@ - (void)updateWithItem:(TKMSubjectModelItem *)item {
...
+      [self.meaningLabel setHidden:NO];
+      self.meaningLabel.text = [self formattedIntervalUntil: [item guruDate] label: @"Guru"];
+    } else if (item.assignment.isLessonStage) {
+      [self.readingLabel setHidden:NO];
+      self.readingLabel.text = [self formattedIntervalUntil: [item guruDate] label: @"Guru"];
+      [self.meaningLabel setHidden:YES];
+    } else {
+      [self.readingLabel setHidden:YES];
+      [self.meaningLabel setHidden:YES];
+    }
```
davidsansome: You'll need to set the font of both labels here too, in case one of these views is reused.
[UIFont systemFontOfSize:kFontSize] should be ok.

indirect:
All done! Let me know if there's anything else. :)

PR #146, opened 2019-07-20
Include audio URLs with the app, and use them

indirect:
This may not be worth merging in the long run, but for now this means the app can play audio for the new subjects that hasn't been uploaded to tsurukame.app yet.


indirect:
Superseded by #155.

PR #171, opened 2019-09-11
Make keyboard cheats use the command key

### LINKED ISSUE #166: Auto-adding empty synonym when using iPad keyboard
Sometimes when typing with my Bluetooth connected keyboard, the app will skip allowing me to input a correct answer and when I explore the item being reviewed, it now has an empty synonym. For instance, I have the reviews set up to test on reading and then meaning:
1. Incorrectly answer reading.
2. Quickly skips the meaning and then goes onto the next item.
3. Reviewing the item on the webpage has an empty synonym.
It’s a little tricky to reproduce, but seems to be correspond to typing too quickly.

indirect:
It's easy to accidentally type a single letter by itself after missing a review, and the cheat actions have no undo. By making these shortcuts command key shortcuts, it's much harder to accidentally trigger them.

Fixes #166, I think.

### DIFF
--- a/ios/ReviewViewController.mm
+++ b/ios/ReviewViewController.mm
@@ -1068,6 +1068,13 @@ class AnimationContext {
   [self markAnswer:TKMOverrideAnswerCorrect];
 }
 
+// For no particularly apparent reason, this seemingly pointless implementation
+// means that holding down the command key after (say) pressing ⌘C does not
+// repeat the action continuously on all subsequent reviews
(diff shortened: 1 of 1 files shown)


indirect:
Oh no, don't merge this yet... holding down the command key causes the cheat to repeat continuously every few hundred milliseconds. 😬

indirect:
This... apparently fixed it? But I don't know why adding an implementation of a method that explicitly returns the same method's super results does that. I guess I'll take it. 😬

davidsansome:
Thanks for fixing this one!

PR #179, opened 2019-10-25
Fix spelling of 'installing' in the 'Offline audio' settings screen.

colons:


davidsansome:
I never noticed that, thank you!

Issue #132, opened 2019-06-05
It runs on macOS 10.15! :D


indirect:
<img width="1280" alt="Screen Shot 2019-06-05 at 12 46 24 PM" src="https://user-images.githubusercontent.com/78/58985770-f52e3f80-8790-11e9-8a60-c85f92b53af4.png">

reviews need a little UI work though 😂 

<img width="901" alt="image" src="https://user-images.githubusercontent.com/78/58985800-037c5b80-8791-11e9-87af-1956c853edd0.png">

indirect:
Update on macOS builds: macOS 10.15 b4 seems to have fixed the last of the alignment bugs, and I should have a build that people can try out soon if they are comfortable running a beta OS known to destroy iCloud data at random. 😅

indirect:
http://files.arko.net/63a4ecbabd0e/Tsurukame.zip

grobmeier:
@indirect This is awesome news. Instead of downloading a ZIP, can I somehow git pull and compile your stuff? I got xcode running, but have no clue on osx development - a readme is appreciated

indirect:
@grobmeier the zip is me compiling the master branch with macOS turned on, there’s nothing to pull.

turtlemaster19:
This seems awesome. This should be added and released in the App Store.

sciencemanx:
I think it would be nice to polish a bit before putting it onto the App Store (add more mac specific layout as makes sense), but maybe the value to people having it now would outweigh the current awkwardness of the iOS layout on mac.

turtlemaster19:
UPDATE: macOS 10.15 is stable now ;)
Maybe we should consider doing this, even without any Mac-specific changes, because the ability to review offline with those scripts, like was stated above, could be useful.

indirect:
Feel free to send a PR!

turtlemaster19:
@indirect PR merged. Can we close this?

turtlemaster19:
Speaking of this, @davidsansome, can you release this to the Mac App Store next update?

Issue #62, opened 2018-10-22
macOS port


nkaretnikov:
Do you have any plans to have a macOS port? If I had my laptop with me and didn't have internet connection, I would rather use it than an iOS app.

davidsansome:
Yes I'd like to do this!

I've never written a native macOS app before though so I've no idea how hard it would be

Rion-Kaneshiro:
Since apple announced it will be possible running iOS apps on the Mac next year, this might be pretty easy to than.
But I’d say it should be an fully adaptive app that looks great on the iPad :)

turtlemaster19:
#132 could solve this issue by using https://developer.apple.com/mac-catalyst/
@davidsansome, should we close this issue or the other? I would love a macOS app, though!

turtlemaster19:
@nkaretnikov PR merged, expect macOS support in the next App Store release (hopefully)

bmironer:
Having a bit of fun experimenting with the macOS version on the master branch.
I'm finding plenty of issues / improvements with usability and layout and even a few crashes. 
Not sure any of my fixes would be deemed clean enough for a pull request though.

Anyway, is anyone still working on the macOS variant? 
If so, is there a more appropriate place to report issues? or maybe a 'macOS' label ?

turtlemaster19:
@bmironer Feel free to create a PR anyway, since it might at least bring attention to the problems. If you feel the need to report an issue somewhere other than here or #132, you can create an issue. There aren’t really others currently working on it.

turtlemaster19:
> Yes I'd like to do this!

@davidsansome Would you consider **officially** releasing the app for the Mac? Building from source, the app already supports it.

matchai:
The app works very well on M1 machines. I'm sure that folks on Intel machines would appreciate the official release of a Universal app 😊

UInt2048:
@matchai Yea, we can build it from source on Intel and it work well there too. 😃

sciencemanx:
I see this is not in the Mac App Store yet. I am not aware if there is other discussion elsewhere, but what needs to be done to release publicly?

UInt2048:
@sciencemanx The macOS checkbox needs to be checked on Sansome's end 😜

bmironer:
And a few annoying bugs with the keyboard shortcuts might be worth fixing too 👍

PR #181, opened 2019-10-28
Add macOS support by checking macOS checkbox

turtlemaster19:
Should solve #62 and #132


turtlemaster19:
Thank you, @davidsansome for adding this. If you haven't already, don't forget to add the Mac development certificate to the development team so it could be released to the Mac App Store.

PR #185, opened 2019-10-31
Study shortcut improvements

indirect:
The main screen keyboard shortcuts for lessons and reviews would wrongly try to trigger even if there were no lessons or reviews, and could get the app into a stuck state.

This also adds a convenience shortcut of "push return to keep studying", which will work as long as there are reviews or lessons remaining.


turtlemaster19:
Should we have return do reviews before lessons instead of the other way around?

PR #186, opened 2019-10-31
Allow reviews without clicking on macOS

indirect:
This is the only patch that seemed really needed to make reviews on macOS work. Without this patch, you have to click the mouse to focus on the text entry box after each animation. With this patch, you can type answers one after another, like on an iPhone or iPad.


turtlemaster19:
#62 and #132

PR #195, opened 2019-11-05
Changed all Swift versions to 5.0

turtlemaster19:
Note that this PR doesn't fix any build errors, so make sure that you can build it before merging. If needed, edit the repository and commit whatever needs to be. If you would prefer to leave it the way it was, close the PR without merging.

This is due to 8180e2daf5b9ae5f8d9628d5e241db0910f2a6e5 upgrading some but not all of the Swift versions set from 4.2 to 5.0.

### DIFF
--- a/ios/TKMDownloadViewController.m
+++ b/ios/TKMDownloadViewController.m
@@ -134,7 +134,7 @@
   if (httpResponse.statusCode != 200) {
     [self reportErrorOnMainThread:filename
                             title:[NSString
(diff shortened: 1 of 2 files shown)


davidsansome:
Thanks, I forgot about the other targets when I made that change.
It's probably better to set the swift version on the whole project and remove the per-target overrides though.  I've done that in 88aaa29cffd9d6ba17ebec90ec02ff7e596c2f44.

turtlemaster19:
I agree. I just wanted to address the issue, and that was the best way I could think of. ;)

Issue #202, opened 2019-11-13
Animate SRS dots on rank up


bmironer:
I love the new little dots for SRS ranking! 
But now it kinda feels there's something missing. 
I think I'm expecting them to 'pop' when I get the answer right (or wrong) and the rank changes.
Maybe a simple animation of the dot filling in / out, or just blinking it once or twice...

davidsansome:
@mzsanford what do you think? 😄

mzsanford:
I'll work on adding some transition animations

turtlemaster19:
If we're going to do this, is #10 a thought?

PR #205, opened 2019-11-13
Animated SRS Level Dots on Level Up  🎉

mzsanford:
As suggested by @bmironer in issue https://github.com/davidsansome/tsurukame/issues/202 I added an animation to the new SRS  level dots on level up. The animation is a modified version of the existing "pop" but re-aligned to look like the dots are popping.

I also tested some transition animations between review items when they are not leveling up. All of them added complexity and given the fast transition and small area of the view they didn't feel worth it. I may spend more time on those in the future.


davidsansome:
This is really cool - thanks again!

Issue #97, opened 2019-02-09
Dark Mode


aoi-otoko:
An optional dark mode would be much easier for night work.

turtlemaster19:
Same as #142

turtlemaster19:
It has now been implemented. Expect to see it in the next update!

turtlemaster19:
@davidsansome, can we close this issue since 216 is tracking remaining issues with dark mode?

PR #211, opened 2019-11-24
Dark mode support

indirect:
There's probably still more that needs to be done, but this at least makes all the text readable in dark mode.

Untitled

### DIFF
--- a/ios/CurrentLevelChartController.m
+++ b/ios/CurrentLevelChartController.m
@@ -142,8 +142,8 @@ static void UnsetAllLabels(ChartViewBase *view) {
   }
 
   PieChartDataSet *dataSet = [[PieChartDataSet alloc] initWithEntries:values];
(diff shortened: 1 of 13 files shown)


indirect:
There are two (subtle-ish) things I can see could be better:

1) The text background colors could be dark-shifted and have white text instead. Or maybe they could just straight up be the regular wanikani colors and have white text?
2) The `NSAttributedString` containing the text for the "Meaning Explanation" and "Reading Explanation" somehow applies the _default_ background color (which in dark mode seems to be black?) to the rest of the "square" after the text is done, to the right hand side of the line. I read a bunch of stack overflow answers and it seems like some people think fixing it is possible, but it wasn't immediately clear to me how to do it.

turtlemaster19:
This would definitely help with #97. Personally, I feel that any dark mode support would be better than none, and this looks great from the screenshot you posted!

PR #223, opened 2019-12-02
Do we still need the 1.2 branch?

turtlemaster19:
It has been years since 1.2 was released. Is there a reason to keep the Release 1.2 branch, or is it just a leftover?


davidsansome:
Is there any reason to delete it?

turtlemaster19:
@davidsansome, is there any reason to keep it? Having useless branches that we're never going to use is just a waste of space for people downloading the repo.

davidsansome:
Ok makes sense.  I've deleted the release-1.2 branch!

Issue #19, opened 2018-04-23
Option to turn off app badging in the settings menu


davidsansome:
Currently users have to do it in the iOS settings app, which is inconvenient.

turtlemaster19:
@davidsansome, it exists now, but I currently have 15 reviews and after turning that off, it still remains.
![3A5FE91A-7016-4A34-ACE5-3EF389636CEB](https://user-images.githubusercontent.com/46784000/69904462-3e400480-136c-11ea-9269-5c75f2406e7f.png)

turtlemaster19:
@davidsansome, doesn't #227 close this?

turtlemaster19:
This issue is now used for the app badging bug as described in #286

bmironer:
seems fixed (commit 182b4cb28d4ba15e54f644f07ae5f101de372589)

turtlemaster19:
This issue can be closed then.

UInt2048:
#227 partially solved the issue - but the count still comes back later...

PR #227, opened 2019-12-03
Clear App Badge When Disabled

mzsanford:
## Problem

If a user disables app badging while a badge is present the current badge value is left in place forever. The `updateAppBadgeCount` method checks the setting and does not update it.

## Fix

Clear the app badge when disabling badging in the settings controller. Fixes a bug in #19


davidsansome:
I'd been meaning to fix this one day - thanks!

Issue #216, opened 2019-11-27
Dark mode checklist


indirect:
Here's a running list of dark/light mode things I've noticed that could be improved. I plan to work through them as I have time, but anyone should free to tackle any of these and open a PR.

Needs a fix in dark mode:
- [x] remaining time (eg 3d 4h) should be light
- [x] "burned" label on dashboard should be light/readable
- [x] review answer field text should be visible as it is typed
- [x] lesson hint text should be dark background, light text
- [x] chart labels text are black but should be white
- [x] multiline strings right hand margin color should not be visible
- [x] hint text on e.g. kanji mnemonics should update to secondary text color
- [x] lessons chip holder should be dark background
- [x] highlight colors should be dark wanikani colors, with white text
- [x] download audio list should be dark background, light text
- [x] font chooser screenshots should be dark background, light text

Needs a fix in light mode:
- [x] NSAttributedString should not have visible backgrounds
- [x] text in Review mode description, above the answer text field

turtlemaster19:
Great! Just referencing #97 in case anyone is following along there

indirect:
Once #229 is merged, the remaining three will be harder:

Highlight colors are NSAttributedStrings set to hard-coded colors, and don't (currently) update when dark mode activates. My theory for how to fix this is to switch to using Asset colors, which have the ability to include dark mode variants. That might mean requiring iOS 10+.

The download audio seems easier but more tedious, updating every color used in those views.

The font screenshots is probably the hardest, since the FontScreenshotter doesn't even build right now. Someday, maybe.

mzsanford:
I think I have an easy fix for the highlight issue, PR upcoming.

davidsansome:
I've done most of these now!  There's one final one I want to fix: the "Meaning" / "Reading" bar above the answer field on the review VC is a bit bright.

turtlemaster19:
> That might mean requiring iOS 10+.

We already do.

PR #228, opened 2019-12-03
Fix Reading Prompt Color on master

mzsanford:
The addition of Dark Mode support broke "light mode" by putting black text in a dark gradient:

_Before_
IMG_7CBC8B0989A7-1

So I fixed that:

_After (light mode)_
Simulator Screen Shot - iPhone 11 Pro - 2019-12-03 at 11 15 44

_After (dark mode)_
Simulator Screen Shot - iPhone 11 Pro - 2019-12-03 at 11 15 57


turtlemaster19:
@indirect, does this conflict with #216?

mzsanford:
I think this checks off "text in Review mode description, above the answer text field" which is already showing checked but appear to be broken in master. The PR for that issue (https://github.com/davidsansome/tsurukame/pull/211) introduced this because of the custom gradient background which is always dark, so the system label color does not contrast. Another possible fix is to change the background gradient in dark mode to be light, but that felt too jarring.

turtlemaster19:
That means @indirect already fixed it in his local copy. If you're reading this, can you submit that stuff?

davidsansome:
Thanks!

PR #229, opened 2019-12-04
Dark mode fixes

indirect:
This fixes the main (and easy) bits from #216.

In dark mode:
- [x] remaining time (eg 3d 4h) should be light
- [x] "burned" label on dashboard should be light/readable
- [x] review answer field text should be visible as it is typed
- [x] lesson hint text should be dark background, light text
- [x] chart labels text are black but should be white
- [x] multiline strings right hand margin color should not be visible
- [x] hint text on e.g. kanji mnemonics should update to secondary text color
- [x] lessons chip holder should be dark background

In light mode:
- [x] NSAttributedString should not have visible backgrounds
- [x] text in Review mode description, above the answer text field


davidsansome:
Amazing, thanks so much for fixing these!

PR #238, opened 2019-12-13
Delete Overrides

### LINKED ISSUE #220: 楓 broken reading information
The Kanji [楓](https://www.wanikani.com/kanji/楓) has a broken reading explanation in the app.
700EFD06-90A2-45DB-BF9B-123148B322F0

turtlemaster19:
Most of these were just bugs with the site. If the local DB was updated, that alone would resolve #220. If any of these are still needed after a DB update, please keep them and close the PR without merging. ;)


turtlemaster19:
Whatever you decide, since [there was another Content Update](https://community.wanikani.com/t/wanikani-content-updates/33117/50), can you refresh the DB and update the app, @davidsansome?

turtlemaster19:
Just like with the 1.2 branch, I am not against keeping it if there is still value in it. It just appears to no longer be necessary anymore, if the site has corrected the problems that this file is overriding in the first place.

turtlemaster19:
P.S. Hope you are doing well wherever you are!

PR #242, opened 2019-12-21
Added newest & oldest available sort (Fixes #173)

### LINKED ISSUE #173: Add Review Order option of "most recently due"
Hi - my use case is having a large backlog of reviews after taking a long time off. WK recommends I don't start over, but just keep hacking on my backlog until I clear it. 

I find myself in a situation where I encounter a card I don't recognize (in some cases effectively having to re-learn it), and it is then scheduled for review in (e.g.) 6 hours. But because of the large number of reviews, I don't see it again for a long time. This makes it hard to consolidate, which I would prefer to prioritize before reviewing further 'new' cards. 

Would it be possible to have an option to sort review order by most recently falling due? I thought Review Order: "SRS level order" might do something like this, but I think that is instead working on the apprentice / guru / master / enlightened / burned hierarchy.

turtlemaster19:
Should fix #173 - attempted to add newest & oldest available sort.


turtlemaster19:
To implement this, the buttons will have to be added to the view controller. ;)

turtlemaster19:
@davidsansome I think it might be ready to merge.

PR #244, opened 2019-12-21
Hide percent for lessons (Fixes #231)

### LINKED ISSUE #231: Hide accuracy during Lessons
It's just a distraction (especially given that no cheating is allowed for typos that it didn't catch). Also, optional hiding it in Reviews would be nice!

turtlemaster19:
To attempt to fix #231, this commit hides the accuracy percentage if it's a lesson.


davidsansome:
Thanks!  I'd prefer to implement this via the delegate though, so I've done it in b2a498f2d5cb389eed7cb3c0e322ed4969915345!

turtlemaster19:
Thank you!

Issue #230, opened 2019-12-04
Feature suggestion: show all readings in kanji


timmy-lee:
Something like this when looking up kanji:
![image](https://user-images.githubusercontent.com/20361890/70157348-477de980-166a-11ea-8e94-3189231b2964.png)

It looks like this only appears when looking up specific kanji and not in the lessons. So as far as this request is concerned, having that feature parity would be nice, but having it appear in the lessons would be even nicer.

timmy-lee:
@davidsansome If this sounds like a feature that you think would be a nice addition, I'd be happy to try tackling it and familiarizing myself with this project to help with future contributions. 😄

turtlemaster19:
While I am not him, I was contemplating this while opening #218. Hopefully this actually is implemented, (but maybe not in Lessons) as this is standard WK (and AlliCrab) behavior

davidsansome:
WaniKani marks these non-primary Kanji readings as wrong, which is why they're not shown during lessons so you don't learn incorrect answers.
I'd be happy to show these other readings outside of lessons.  We could maybe show them during lessons as well if we made it very clear which one is the "correct" Kanji reading, and that the other one is only being shown for information purposes and not as part of the lesson.

The code that hides non-primary readings is here: https://github.com/davidsansome/tsurukame/blob/master/ios/SubjectDetailsView.swift#L74 and https://github.com/davidsansome/tsurukame/blob/master/ios/SubjectDetailsView.swift#L142.  You'd probably need to add a showNonPrimaryReadings bool to SubjectDetailsViewController (https://github.com/davidsansome/tsurukame/blob/master/ios/SubjectDetailsViewController.m#L43) which we can set to false in lessons and true elsewhere.

Do you want to have a go at implementing that?  I can answer any more questions if you have them!

timmy-lee:
By enabling the primary readings in the way you mentioned, I get something like this:
![image](https://user-images.githubusercontent.com/20361890/70448096-b2596700-1a54-11ea-9c9f-edf917c04d62.png)

But I was thinking more of something like this (the readings themselves haven't been sorted yet):
![image](https://user-images.githubusercontent.com/20361890/70448379-21cf5680-1a55-11ea-8452-f4efb0befab5.png)
While having the non-primary reading(s) in light gray text or having the primary reading in bold text. What do you think @davidsansome?

sciencemanx:
[...]

PR #245, opened 2019-12-23
Add Option to Show All Readings

timmy-lee:
Implements https://github.com/davidsansome/tsurukame/issues/230
* Unless requested, I opted not to make separate sections for each type of reading as @sciencemanx brought up a good point about it potentially taking up too much room before getting to the "meat" of the entry/subject.
* Primary readings are always first. If more that one reading exists, the primary one(s) will be in bold.

### When option is flipped on:
* Kanji/Vocabulary with multiple readings

* Kanji/Vocabulary with one reading (functions the same as today)

* No changes in radicals

### When option is off:
* No changes


turtlemaster19:
Looks great! Happy Holidays!

davidsansome:
Awesome, thanks for implementing this!

PR #248, opened 2019-12-26
Pause 30 seconds if too many requests (Fixes #233)

### LINKED ISSUE #233: Error 429 - Too Many Requests
While trying to generate the data.bin, some of the radicals aren't downloaded because of this error.
HTTP429

Another reason we should consider using the API to get data (#180) and converting Go code to Swift (#189)

turtlemaster19:
Pauses 30 seconds if HTTP error 429: Too Many Requests occurs, to fix #233.

### DIFF
--- a/cmd/scrape/main.go
+++ b/cmd/scrape/main.go
@@ -17,6 +17,8 @@ package main
 import (
 	"flag"
 	"log"
+	"time"
+	"strings"
 
 	"github.com/davidsansome/tsurukame/api"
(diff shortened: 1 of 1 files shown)


davidsansome:
I'd rather throttle requests to 1qps to stay under the rate limit (https://docs.api.wanikani.com/20170710/#rate-limit) than add an arbitrary pause

turtlemaster19:
Great! I didn’t exactly like the 30 seconds anyway.

Issue #213, opened 2019-11-25
Don't allow kanji in reading


turtlemaster19:
If the user uses a Japanese IME and types in a kanji in the reading, it shouldn't be allowed. For example, for two, the katakana should still be able to submit, but the kanji (which would be marked wrong) should not. This would match WaniKani behavior.

PR #252, opened 2020-01-06
Narrow accepted character range (Fixes #187 and #213)

### LINKED ISSUE #187: Don't allow Japanese in meaning
If in a review, such as 一つ, the user types ひとつ, or any other Japanese characters, for the meaning, don't allow it to be submitted.

turtlemaster19:
To fix #187 and #213, this code attempts to narrow the valid character range.

### DIFF
--- a/ios/AnswerChecker.swift
+++ b/ios/AnswerChecker.swift
@@ -24,13 +24,13 @@ import Foundation
   }
 
   @objc static let kAsciiCharacterSet = CharacterSet(charactersIn: Unicode.Scalar(0x00) ..< Unicode.Scalar(0x7F)!)
-  @objc static let kKanaCharacterSet = CharacterSet(charactersIn: Unicode.Scalar(0x3040) ..< Unicode.Scalar(0x309D)!)
-  @objc static let kAllKanaCharacterSet = CharacterSet(charactersIn: Unicode.Scalar(0x3040) ..< Unicode.Scalar(0x3100)!)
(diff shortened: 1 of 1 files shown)


ios/AnswerChecker.swift
```diff
@@ -23,24 +23,27 @@ import Foundation
...
-      "さしすせそざじずぜぞ" +
-      "たちつてとだぢづでど" +
-      "なにぬねの" +
-      "はひふへほばびぶべぼぱぴぷぺぽ" +
-      "まみむめも" +
-      "らりるれろ" +
-      "やゆよゃゅょぃっ" +
-      "わをん")
+  @objc static let kAsciiCharacterSet = CharacterSet(charactersIn: Unicode.Scalar(0x00) ..< Unicode.Scalar(0x7F)!)
+  @objc static let kKanaCharacterSet = CharacterSet(charactersIn: Unicode.Scalar(UInt32(0x3040))! ..< Unicode.Scalar(UInt32(0x309D))!)
```
davidsansome: This should be called kHiraganaCharacterSet?

turtlemaster19: Agree. I didn't want to change your name. Will change it, though.

turtlemaster19:
@davidsansome Ready to merge.

ios/AnswerChecker.swift
davidsansome: Oops that was my fault, I didn't notice :smile:  Thanks!

PR #256, opened 2020-01-09
Update README to address CocoaPods & signing (Fixes #249)

### LINKED ISSUE #249: Update README to address how to install dependencies
Since the dependencies aren’t included in the app (for good reasons), and come from different sources, the README should explain how to install the dependencies that need to be installed.

turtlemaster19:
Fix #249 by updating README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,46 @@
 [![CircleCI](https://circleci.com/gh/davidsansome/tsurukame.svg?style=shield)](https://circleci.com/gh/davidsansome/tsurukame)
 
-Tsurukame is an unofficial WaniKani app for iOS.  It helps you learn Japanese
-Kanji.
+Tsurukame is an unofficial WaniKani app for iOS.  It helps you learn Japanese Kanji.
 
 You can download Tsurukame on the App Store.
 
 [![Download on the App Store](https://devimages-cdn.apple.com/app-store/marketing/guidelines/images/badge-example-preferred.png)](https://itunes.apple.com/us/app/tsurukame-for-wanikani/id1367114761)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,17 +1,46 @@
...
-## Using the tools
+## Install the CocoaPods dependencies
+First, you will need to install [CocoaPods](https://cocoapods.org) if you don't already have it. The website suggests using `sudo gem install cocoapods`.
+
+Next, you will need to switch the current directory to the `ios` directory. You can do this using:
+
+	cd /path/to/tsurukame/ios
+
+Of course, replace `/path/to` with the actual path to the Tsurukame directory.
+Finally, simply install the dependencies using `pod install`.
```
davidsansome: I really don't think you need to run `pod install`.  And I don't think you need to even install cocoapods.  You should only need these if you change or want to update the dependencies.

Did you try building without running `pod install` and get an error?

README.md
```diff
@@ -1,17 +1,46 @@
...
+
+## Change the signing identifiers
```
davidsansome: Thanks :)
I want to configure CircleCI to automatically sign builds and put them on TestFlight, so this section might change soon, but looks good for now.

turtlemaster19: That would be amazing! You will have to share the TestFlight link on the forums, since I imagine there are quite a few people that don't know how to build themselves (or can't) but want to beta test.

README.md
turtlemaster19: I suppose it works. I thought it should be included, but I guess I forgot it could be built without this step.

PR #258, opened 2020-01-14
Remove Apprentice level-up popup (Fixes #255)

### LINKED ISSUE #255: Remove Apprentice level-up popup (Revert #6)
Now that #224 has been implemented, we need to correct #6. I see a few options:

1. Revert #224 (that seems ridiculous though)
2. Don't show the Apprentice level-up at all (an item can't down-level past Apprentice I)
3. Always show the level-up when an item increases (Apprentice II, Guru II, etc.)
4. Have an option to choose between 2 and 3

~~Of these options, it seems that the fourth option is probably the best.~~
**Edit:** It seems that the second option has been chosen.

turtlemaster19:
Fix #255 by removing the Apprentice level-up popup.

### DIFF
--- a/ios/LessonsViewController.m
+++ b/ios/LessonsViewController.m
@@ -189,7 +189,7 @@
 
 #pragma mark - ReviewViewControllerDelegate
 
-- (bool)reviewViewControllerAllowsCheatsForReviewItem:(ReviewItem *)reviewItem {
+- (BOOL)reviewViewControllerAllowsCheatsForReviewItem:(ReviewItem *)reviewItem {
   return false;
 }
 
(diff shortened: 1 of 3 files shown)


turtlemaster19:
@davidsansome, ready for review.

davidsansome:
I'm not really sure I want to add an option for this - it seems like a very minor setting that we should probably make a decision one way or the other for the user.

turtlemaster19:
Okay. Which one should it be? I needed to fix this now that SRS stage 2 is now named Apprentice II instead of just Apprentice, so we shouldn't have a level-up popup for this stage, unless we use the detailed name.

davidsansome:
Let's keep the current behaviour but just remove the Apprentice popup!

turtlemaster19:
Sounds good. I'll go do that.

turtlemaster19:
@davidsansome Ready for review

PR #259, opened 2020-01-14
Use SwiftPM instead of third_party for Rexy

turtlemaster19:
The `third_party` directory can be removed now, since Rexy is able to be installed via Swift Package Manager in Xcode. This commit does that, removing the need to keep a local copy of Rexy, since it will automatically be installed at build time.


davidsansome:
I made a bunch of local UTF-8 fixes to Rexy that I didn't upstream yet (since the project looked kinda dead).  They'd need to be upstreamed before we can move this out of third_party.

turtlemaster19:
@davidsansome, were they all in `Regex.swift`?

davidsansome:
Yep I think so!  You should be able to diff it against the upstream one to see.

turtlemaster19:
If that's the case, I already took care of them when https://github.com/vadymmarkov/Rexy/pull/14 was merged.

davidsansome:
Ouch, I guess he didn't review that very well then - there were some breaking API changes in there too...

turtlemaster19:
Oops, I figured they were needed to keep it working, since you made some fixes there that were required. Should I try to revert them, or should we leave it?

P.S. I'm not sure why CircleCI is failing, since it worked perfectly fine building locally.

davidsansome:
You don’t have permission to save the file “repositories” in the folder “SourcePackages”." UserInfo={NSFilePath=/Users/SourcePackages/repositories

I think it's a problem with the CircleCI config - it looks like I didn't set a path that it needs somewhere.  I'll try to fix it tomorrow!

With Rexy, the biggest change was changing the return type of all the methods from String to Substring (so the caller can see the position that was matched).  It would break anybody using the package that expected them to be Strings.

turtlemaster19:
I'm not sure on whose side it is. It looks like https://github.com/fastlane/fastlane/issues/15454

turtlemaster19:
@davidsansome It seems they were aware, since they numbered the version 1.0 to signal the breaking changes. I resubmitted the commit to incorporate the latest changes in master. (Still failing. 😢)

turtlemaster19:
@davidsansome Error seems to be cause by Fastlane running a command with `-derivedDataPath '/Users'`

turtlemaster19:
@davidsansome Error seems to be cause by Fastlane running a command with `-derivedDataPath '/Users'`

turtlemaster19:
@davidsansome Fastlane seems to be convinced that Xcode should use `/Users` as the `DerivedData` folder. Any luck with this?

davidsansome:
Sorry I didn't get time to look at it yet!

turtlemaster19:
@davidsansome I see you with the TestFlight stuff. 👀 It's definitely better 1-2 days.

turtlemaster19:
Problem seems to be fastlane/fastlane#15910

turtlemaster19:
Why do I keep accidentally closing?

turtlemaster19:
@davidsansome It seems to still be trying to find the old Rexy. Do you know why?

turtlemaster19:
@davidsansome, ready for review.

turtlemaster19:
You may need to rollback to 09848b6 if you're ready before I can fix this.

turtlemaster19:
Going to merge the changes just committed to master. Give me a few moments.

turtlemaster19:
@davidsansome Ready for review. Tests are failing for the same reason master is - which is beyond the scope of this PR

davidsansome:
Thanks!  Sorry again for taking my time with this one.
I'm looking at why the master tests are failing now...

PR #266, opened 2020-01-16
Fix crash on login for some new users

### LINKED ISSUE #263: App crashed on sign-in!
Hiya, I just tried to log in and the app crashed immediately. It won't re-open either.
I'm using iOS 13.1.2 on an iPhone 6s for reference.

Thank you for the app, so excited to use it!! :)

mzsanford:
There is some case where the code has `level == 0` and the guard (`<= 0`) was not working as intended. Change what was effectively `<= 0 && > max` to `<= 0 || > max`, fixes #263.


turtlemaster19:
@mzsanford Is this a joke? Levels start at 1.

turtlemaster19:
I created a forum topic: https://community.wanikani.com/t/should-we-have-a-level-0/41419

mzsanford:
My crash reproduction on fresh login with my test account:

> 2020-01-16 10:24:07.142171-0800 Tsurukame[22551:2419823] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM objectAtIndexedSubscript:]: index 18446744073709551615 beyond bounds [0 .. 59]'

and here is the debugger:

CleanShot 2020-01-16 at 10 35 08@2x


If things worked like we thought there wouldn't be crashes like this 😉

turtlemaster19:
Ok. Did you mean Level 1 users, then?

mzsanford:
This could have to do with when the database is initialized on first login, maybe before we know the level in some cases. All I know is that old if was not doing what it looked like and this fixed it.

DsfufGe

turtlemaster19:
@mzsanford Please fix the title before it gets noticed.

davidsansome:
Thanks so much for fixing this so quickly!
I've got lots of reports about this since releasing 1.16 yesterday.

PR #267, opened 2020-01-16
Auto switch keyboard to default, not 'en'

### LINKED ISSUE #264: Meaning keyboard always reverts to QWERTY.
since #36 was released, the keyboard now reverts to English QWERTY on every meaning question, instead of staying on the last selected layout, which is a real pain if you're not used to it.
Note : I'm using the French AZERTY layout, but my OS localization is in English-International.
This happens regardless of the wanikani setting.

mzsanford:
Fixes #264


turtlemaster19:
Great! For the record, tests are failing due to #261 not being merged yet

Issue #83, opened 2018-12-11
Reviews from higher levels shown


hesch:
The app shows reviews for levels higher than your current level. After the content overhaul some items were moved to higher levels. These items can't be reviewed on the website, but tsurukame shows them for review. I asked the WaniKani-Support if this is a bug on their side, but they said it was intentional and the reviews would resume when the user reaches the new level of the item.

To make the case more clear here is an example:
- I am level 30 and do a lesson for an item, that is in level 30 at this moment
- The item gets moved to let's say level 35
- After 4 hours the review shows up in Tsurukame but not on WaniKani

I'm not entirely sure this should be fixed (I mean why shouldn't you be able to review already learned items?), but it is a inconsistency between the website and the app.

onimitch:
Maybe it could be offered as an option in settings about whether to include higher level items

PR #269, opened 2020-01-17
Skip reviews for items that have moved to a higher level on the backend

mzsanford:
### 久しぶり Bug #83 

As reported in #83 and more recently [on the forum](https://community.wanikani.com/t/ios-tsurukame-native-app-with-offline-lessons-and-reviews/30635/505). Some items were moved to higher levels during a Wanikani content update, for example "久しぶり". The WK API does not accept the reviews for those but Tsurukame keeps rescheduling them since the state does not change. The results in an item showing up for review every 4 hours until you reach the level to which it has moved. In my case the estimate is some time in early 2021.

This branch filters out any assignment with a level higher than your current level. I'm not sure a settings toggle makes sense because the API does not accept it. If someone has a use case for that I can add one.


turtlemaster19:
Sounds good! Nice branch name. :eyes:

ios/LocalCachingClient.mm
```diff
@@ -657,13 +657,21 @@ - (void)maybeUpdateAvailableSubjectCounts {
...
+  TKMUser *userInfo = [self getUserInfo];
   for (TKMAssignment *assignment in assignments) {
     // Don't count assignments with invalid subjects.  This includes assignments for levels higher
-    // than the user's max level.
+    // than the user's max subscription level.
     if (![_dataLoader isValidSubjectID:assignment.subjectId]) {
       continue;
     }
 
+    // Skip assignments that are a higher level thn the user's current level. Wanikani items that have
```
davidsansome: Typo: s/thn/than/

ios/ReviewItem.m
```diff
@@ -20,13 +20,18 @@
...
-                                          dataLoader:(DataLoader *)dataLoader {
+                                          dataLoader:(DataLoader *)dataLoader
+                                  localCachingClient:(LocalCachingClient *)localCachingClient {
   NSMutableArray *ret = [NSMutableArray array];
   for (TKMAssignment *assignment in assignments) {
     if (![dataLoader isValidSubjectID:assignment.subjectId]) {
       continue;
     }
 
+    if (localCachingClient.getUserInfo.hasLevel && localCachingClient.getUserInfo.level < assignment.level) {
```
davidsansome: getUserInfo makes a database lookup so can you store it in a local variable so it only does one?

    TKMUserInfo *userInfo = [localCachingClient getUserInfo];
    if (userInfo.hasLevel && userInfo.level ...)

ios/LocalCachingClient.mm
mzsanford: Fixed

ios/ReviewItem.m
mzsanford: I added the `getUserInfo` call before the outer loop and merged in master as well.

turtlemaster19:
@davidsansome Can you close #83 now?

Issue #60, opened 2018-10-15
Siri Integration


hesch:
It would be nice to have an Integration with Siri Shortcuts.
The following Actions would be useful for me:
 - Open Lessons
 - Open Reviews
 - Get Lesson count
 - Get Review count

turtlemaster19:
Great idea. Especially since the Shortcuts app is now built in to iOS 13.

PR #270, opened 2020-01-17
Add Limited Siri Shortcut Support

mzsanford:
This does half of the items in #60 but probably 10% of the work. This adds the ability to launch the app into lessons or reviews from a shortcut. It records with Siri each time those screens are open to help build weighting as well.

The other two items, which would not launch the app, require an extension process and some way to share the data between the main app and the extension. The simplest is probably adding "App Groups" as a capability and then moving the existing cache database into the app group. This would make the offline cache available to extension processes like this (and others, like a today widget). That's a bigger change that require App Store Connect changes from @davidsansome  so I wanted to get this out first and see if it can solve 90% of the problem with 10% of the work.


ios/AppDelegate.m
```diff
@@ -64,6 +64,17 @@ - (BOOL)application:(UIApplication *)application
   return YES;
 }
 
+- (BOOL)application:(UIApplication *)application willContinueUserActivityWithType:(NSString *)userActivityType {
+    if ([userActivityType isEqualToString:SiriShortcutHelper.ShortcutTypeReviews]) {
```
davidsansome: I think isEqual: is faster if you know both of them are strings.

ios/AppDelegate.m
```diff
@@ -64,6 +64,17 @@ - (BOOL)application:(UIApplication *)application
...
+        MainViewController *mainVC = _navigationController.viewControllers[0];
```
davidsansome: Will the top view controller always be a MainViewController?

ios/AppDelegate.m
mzsanford: Done

ios/AppDelegate.m
mzsanford: It was in all of my testing, but to be sure I added a quick loop to check all child VCs.

turtlemaster19:
@davidsansome Can you close #60 now?

davidsansome:
> @davidsansome Can you close #60 now? Or, do you feel that it should remain open?

There's still some more to that bug that this PR didn't cover!

PR #273, opened 2020-01-19
Allow Japanese in search 

### LINKED ISSUE #272: Kanji Search No longer working
I used to be able to search for words using English, romanji, hiragana, or Kanji characters but recently the last two suddenly stopped working. It was very convenient so I hope it was just a bug and that feature is not being removed.

turtlemaster19:
Fix #272 by removing lowercase string check.

### DIFF
--- a/ios/SearchResultViewController.m
+++ b/ios/SearchResultViewController.m
@@ -139,7 +139,7 @@ static bool SubjectMatchesQueryExactly(TKMSubject *subject, NSString *query, NSS
 
     dispatch_async(dispatch_get_main_queue(), ^{
       // If the query text changed since we started, don't update the list.
-      if (query != [searchController.searchBar.text lowercaseString]) {
+      if (query != searchController.searchBar.text) {
         return;
       }
       


turtlemaster19:
I'm not sure whether this actually fixes this, but it's the only thing that has been changed recently enough relating to the search items, that I know of.

davidsansome:
Close!  But it looks like the bug was actually because I used == to compare strings (which actually compares their pointers) rather than using isEqual.

dmikhalsky:
still doesn't work

turtlemaster19:
@dmikhalsky 1.18 hasn't been uploaded to the App Store yet! 😉

dmikhalsky:
> @dmikhalsky 1.18 hasn't been uploaded to the App Store yet!

OK :-) If I Have TestFlight installed, can I become a beta tester?

davidsansome:
@dmikhalsky yes!  You can join the beta here: https://testflight.apple.com/join/Fijye2AA

PR #278, opened 2020-01-22
Change the custom font control gestures to be clearer

### LINKED ISSUE #271: Kanji display area no longer allows a single tap to reveal default font
Since some recent app update, if you have multiple fonts turned on, tapping in the kanji display area no longer temporarily shows the kanji in the standard font (this is handy if the displayed font is really messy and you can't read it) -- now it seems you have to tap and do a short drag either up or down to show it. It's a very subtle difference, but being able to tap (without having to drag a few pixels) to show the standard font was a lot easier.

mzsanford:
Instead of using a long-press, which was hard to discover and use, this moves to a single tap method. One tap to toggle custom fonts, but move back to a random font for the next item. This also changes the swipe gesture to cycle through fonts in a predictable order and allows swiping backward through that order.

Fixes #271


davidsansome:
Unfortunately fontsThatCanRenderText doesn't return the fonts in the same order every time, because it's iterating over Settings.selectedFonts which is an NSSet, and sets aren't ordered.

How about calling fontsThatCanRenderText only once per subject, maybe caching it in randomFont(), then using that cached copy of the [String] rather than calculating it fresh every time?

bmironer:
On that topic : it works but it's really not usable in the macOS version.
Maybe add a keyboard equivalent?
Maybe pressing shift to toggle between default and custom, and left/right arrow to swipe through the list ?

turtlemaster19:
@mzsanford Could that be added? If the macOS build is actually released to the Mac App Store, that would probably be a complaint! ;)

Issue #286, opened 2020-01-31
macOS : various issues


bmironer:
Quick grab-bag of issues in the (home-built) macOS version. 

- all TableViews in the macOS version use grey cell background with white (or black) section header backgrounds. This is the reverse than on iOS and since text still has its white (or black) background, it shows (in settings, kanji info, lessons etc..)

- The labels and text feel small on a 15" macbook screen. 
Maybe globally bump it by 150 or 200% ?
Same for the squarish kanji icons and little page markers in lessons. 
(the 'previous kanji' button in reviews is bigger and feels the right size)

- the dashboard header is missing a bit of inset space at the top.

- the dashboard piecharts do not live-relayout like the rest of the dashboard when dragging the window. They do relayout properly to the new size when coming back from another view.

- macOS build can't verify 3rd party libs (requires disabling library verification to launch)

- no way to dismiss the review ending summary with the keyboard. Is there a standard keyboard equivalent to the 'Done' button? (forgot to try Esc. Return seems obvious but didn't work)

- Pressing 'Down Arrow' on the home screen with no review queued goes to an empty black screen with no way back except quitting. (With reviews queued it goes to reviews)

### Fixed

- in the Audio download screen  'Download audio to your *phone*..." should probably say '... to your device' instead. (fixed by #287)

- badging the icon doesn't work. see #19 (fixed)

- macOS build tries and fails to package the complication module, which is not supported on macOS (fixed)

- the particle 'pop' animation has to finish before the keyboard is active.
Using back-to-back reviews I can start typing the reading right after I press return on the meaning. I then lose the first keystrokes while the animation plays.
Couldn't reproduce on iOS with the on screen keyboard. Maybe the keys are buffered? Didn't try with a bluetooth keyboard either (fixed by #381)

- Answer field loses keyboard focus after 'shake' animation (after a correct kanji answer to a vocabulary question or after a typo on hiragana). (fixed by #381)

UInt2048:
@bmironer Can you add #381 to your fixed list?

bmironer:
> @bmironer Can you add #381 to your fixed list?

sure, I will check the fix later when I get a few more reviews. 

maybe related:  I just found that pressing 'down arrow' on the home screen of the mac version opens an empty black scene with no way back ( commit cfa3dc5aa0c7c0e91a15deb020ba350666ff7eae )

UInt2048:
> macOS build can't verify 3rd party libs (requires disabling library verification to launch)

[...]

PR #287, opened 2020-01-31
Correct audio download text

turtlemaster19:
Fixes audio download text, replacing phone with actual device. Also assists #286

### DIFF
--- a/ios/TKMOfflineAudioViewController.mm
+++ b/ios/TKMOfflineAudioViewController.mm
@@ -93,8 +93,8 @@ static NSData *DecompressLZFSE(NSData *compressedData) {
 - (void)populateModel:(TKMMutableTableModel *)model {
   [model addSection:@""
              footer:
-                 @"Download audio to your phone so it plays without delay and "
-                  "is available when you're not connected to the internet."];
+                 [NSString stringWithFormat: @"Download audio to your %@ so it plays without delay online and "
+                  "it's available when you're not connected to the Internet.", UIDevice.currentDevice.model]];
 
   for (const AvailablePackage &package : kAvailablePackages) {
     TKMDownloadModelItem *item = [[TKMDownloadModelItem alloc] initWithFilename:package.filename


ios/TKMOfflineAudioViewController.mm
```diff
@@ -93,8 +93,8 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
 - (void)populateModel:(TKMMutableTableModel *)model {
   [model addSection:@""
              footer:
-                 @"Download audio to your phone so it plays without delay and "
-                  "is available when you're not connected to the internet."];
+                 @"Download audio to your device so it plays without delay online "
```
davidsansome: What about using UIDevice.currentDevice.model instead of "device"?  It'll say "iPhone" or "iPad".
I use it already in https://github.com/davidsansome/tsurukame/blob/42ef188ea79e223e797a7d62528cc62bf27c61b0/ios/SettingsViewController.m#L369

turtlemaster19: Sounds good!

ios/TKMOfflineAudioViewController.mm
```diff
@@ -93,8 +93,8 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder {
 - (void)populateModel:(TKMMutableTableModel *)model {
   [model addSection:@""
              footer:
-                 @"Download audio to your phone so it plays without delay and "
-                  "is available when you're not connected to the internet."];
+                 [NSString stringWithFormat: @"%@%@%@", @"Download audio to your ", UIDevice.currentDevice.model,
```
davidsansome: This is a bit weird - you'd typically put all the fixed text in the first format string argument, and then only have one %@, like:

    [NSString stringWithFormat:@"Download audio to your %@ so it plays without delay ...", UIDevice.currentDevice.model]

turtlemaster19: No kidding, that would seem weird in a `printf`! I'll fix that. 🤦‍♂

Issue #119, opened 2019-05-07
Apple Watch support


waltiz:
It would be great to have Apple Watch support. 
This would consist in an Apple Watch complication that shows a different kanji every now and then, maybe based on the wall of shame or some other criteria.

dysnomian:
I would love to have a count of my available reviews as a complication.

mzsanford:
I recently did a profession project based on iOS app extensions so I may take a crack at this. The options for complications have some serious design limitations, as your can see in the [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/watchos/app-architecture/complications/). I've started a little bit of testing to see how often it can refresh and such.

Some things that could be included, depending on what's interesting to people:

  - Review Count (**small complication**: current count; **medium size**: current, next hour; **large**: current, next hour, next day)
    - If current review count is 0, maybe time of next review date/time instead
  - Level (current level number, gauge showing progress within level)
  - Lesson Count (current number available)
  - Random Item - the original request.

Those are in roughly the order they seem useful to _me_, but comment if you have other thoughts (@dysnomian , @waltiz ). The way complications are structured the app can only present one. Either (A) a single type needs to be chosen, (B) the type is chosen can be driven by the complication type (large is review count, small is level, etc), or (C) I need to add a settings screen to choose which. Option (C) is the only one that adds significant work.

The original request was for a random item but I feel the small space will make that difficult. It also has the downside of subverting the Wanikani SRS goals. Having said that I'll test it out

mzsanford:
Just in case anyone is following along, my current tests are below using real data transferred from the phone app. Allowing untethered data updates is much more complicated so I've pushed that off until later. Note:

  * The "Reviews" complications with a progress bar are currently fake data. I'm not sure a progress bar makes sense there.
  * The "Level 13.5" is because I am stuck in vocab with an empty progress bar (See also: [WK Community Topic](https://community.wanikani.com/t/leveling-up-before-my-time/38085/7?u=mzsanford))
  * Tapping on the complication launches the Settings screen, which includes a handy debug thing showing how long since the data was updated.

_Settings - Review Counts (default)_
![Simulator Screen Shot - Apple Watch Series 5 - 44mm - 2019-11-20 at 12 55 30](https://user-images.githubusercontent.com/33917/69277699-92d9c780-0b95-11ea-8552-8ed5bbe0bb04.png)



_Review Count Complications_
[...]

PR #289, opened 2020-01-31
Apple Watch Complication

mzsanford:
This adds support for an Apple Watch complication as discussed in #119. It's a bit of a monster review but unfortunately adding a new embedded app is a bit of an all-or-nothing.

## Things to Note

1. The watch app requires a user to use the phone app. The watch app is getting data from the phone app via the complication updates API. Whenever the app loads new data or is closing it posts the latest known good state. To support watch-only installations the code base would need to be refactored heavily to make the data access into a shared framework. That felt beyond what I could do as a contributor.
2. I needed some smaller icons and tried to reuse the existing AppIcon where I could. For the smallest size (`miniCrab`) the  existing app icon was too busy. You only get 24x24 so I created a simpler icon but stayed with a turtle to match the main app. Earlier beta tests used a crab, which has since been removed.
3. This exposes the "half level" method also discussed in #119.
4. Some caode exists in the complication to display random characters on the watch (`templateForCharacter`). There is no path where this is called currently. That was  feature requested in #119 that is incomplete. I'm not sure if it should be finished or removed, honestly. Left here so we can decide.

## Future Ideas

I wanted to get this out as-is for the simplest version. If this is useful I'll be around to fix bugs that happen and finish/remove `templateForCharacter` as needed. After that I may attempt to refactor the complication into a shared component and two complications, "Tsurukame Reviews" and "Tsusukame Level" so user can have access to both.


turtlemaster19:
@mzsanford I noticed some spaces in the file paths. It seems that this is undesirable if you look at other places in the repo. Perhaps instead of `Tsurukame Complication Extension`, we could just use `Complication` or `watch_complication`?

mzsanford:
@turtlemaster19 good point. I renamed to remove the spaces … XCode makes that more complicated than seem reasonable so it took me some time to get to it.

ios/ComplicationExtension/ComplicationController.swift
```diff
@@ -0,0 +1,598 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import ClockKit
+import os
+
+class ComplicationController: NSObject, CLKComplicationDataSource {
+  let TsurukameHighlightColor = UIColor.red
```
davidsansome: These constants should start with a lower-case letter, like `tsurukameHighlightColor`

ios/ComplicationExtension/ComplicationController.swift
```diff
@@ -0,0 +1,598 @@
...
+      handler(entry)
+    } else {
+      handler(nil)
+    }
+  }
+
+  func getTimelineEntries(for _: CLKComplication, before _: Date, limit _: Int,
+                          withHandler handler: @escaping ([CLKComplicationTimelineEntry]?)
+                            -> Void) {
+    // Call the handler with the timeline entries prior to the given date
```
davidsansome: I was confused by this for a second.  Maybe reword this to be `// Don't provide any timeline entries in the past`

ios/ComplicationExtension/DataManager.swift
```diff
@@ -0,0 +1,107 @@
...
+protocol DataManagerDelegate: AnyObject {
+  func onDataUpdated(data: UserData, dataSource: ComplicationDataSource)
+}
+
+enum ComplicationDataSource: Int {
+  case ReviewCounts, Level, Character
+}
+
+class DataManager {
+  private let UserDefaultsKeyData = "LastKnownData"
```
davidsansome: Same with these

ios/ComplicationExtension/DataManager.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return true
+  }
+
+  func addDelegate(_ delegate: DataManagerDelegate) {
+    delegates.append(delegate)
+  }
+
+  func removeDelegate(_ delegate: DataManagerDelegate) {
+    delegates.removeAll { (d) -> Bool in
```
davidsansome: I think this can be:

    delegates.removeAll { $0 === delegate }

ios/ComplicationExtension/SettingsController.swift
```diff
@@ -0,0 +1,108 @@
...
+    (.ReviewCounts, "Review Counts"),
+    (.Level, "Level"),
+    // Removed pending real data
+    // (.Character, "Kanji"),
+  ]
+
+  override func awake(withContext context: Any?) {
+    super.awake(withContext: context)
+
+    os_log("MZS - Awake with context: %{public}@", String(describing: context))
```
davidsansome: Delete this line 😄

ios/ComplicationExtension/SettingsController.swift
```diff
@@ -0,0 +1,108 @@
...
+    for (idx, (dataSource, _)) in dataSourceOptions.enumerated() {
+      if dataSource == DataManager.sharedInstance.dataSource {
+        dataSourcePicker.setSelectedItemIndex(idx)
+      }
+    }
+  }
+
+  override func willActivate() {
+    // TODO: How to get context from ExtensionDelegate#handleUserActivity ?
+    os_log("MZS - will activate")
```
davidsansome: This one too

ios/WatchHelper.swift
```diff
@@ -0,0 +1,202 @@
...
+import WatchConnectivity
+#if os(watchOS)
+  import ClockKit
+#endif
+
+typealias ClientDelegateCallback = (([String: Any]) -> Void)
+typealias EpochTimeInt = Int64
+
+class WatchConnectionServerDelegate: NSObject, WCSessionDelegate {
+  override init() {
```
davidsansome: I don't think you need this initialiser if all it's doing is calling the superclass's

ios/WatchHelper.swift
```diff
@@ -0,0 +1,202 @@
...
+
+  #if os(iOS)
+    func sessionDidBecomeInactive(_: WCSession) {}
+
+    func sessionDidDeactivate(_: WCSession) {}
+  #endif
+}
+
+@objc class WatchHelper: NSObject {
+  public static let KeyReviewCount = "reviewCount"
```
davidsansome: These too

ios/WatchHelper.swift
```diff
@@ -0,0 +1,202 @@
...
+  let DuplicatePacketWindow = TimeInterval(30)
+
+  override init() {
+    super.init()
+    #if os(iOS)
+      startServerSession()
+    #endif
+  }
+
+  @objc static func sharedInstance() -> WatchHelper {
```
davidsansome: Could this be a private(set) public(get) class variable?  I forget the exact syntax.

ios/ComplicationExtension/ComplicationController.swift
mzsanford: Renamed all constants as suggested. Guess the language-hopping catches up with me sometimes :)

ios/ComplicationExtension/DataManager.swift
mzsanford: Nice!

ios/ComplicationExtension/SettingsController.swift
mzsanford: Deleted this and a few others

ios/WatchHelper.swift
mzsanford: Changed to a `static let` which accomplishes that

mzsanford:
@davidsansome Thanks for taking the time to review, I know it was a bit of a monster PR.

PR #290, opened 2020-02-04
Fix crash at startup (Fixes #283)

### LINKED ISSUE #283: macOS: crash at startup once logged in
the `headerView` is sometimes nil in `refresh()` and `updateUserInfo()` at startup with login info in the settings.
`guard let headerView = headerView else { return }` in both functions is enough to workaround.

turtlemaster19:
Fix #283 by adding guard in `refresh(quick:)` and `updateUserInfo()`


turtlemaster19:
@mzsanford Good idea. I think I can close this and we'll move the open PR to #293

PR #293, opened 2020-02-07
macOS: Fix problems in custom font support

### LINKED ISSUE #283: macOS: crash at startup once logged in

### LINKED ISSUE #284: macOS: the font preview resizes improperly
It resizes to the width of the cell, keeping the aspect ratio, which zooms on the center of the image. It should probably not resize at all, or resize the cell with it so the image is still fully visible.

mzsanford:
This branch fixes the image scaling in the font download screen (confirmed it's still ok on the phone), fixes #284. I also added keyboard support (⬅️  & ➡️ ) for cycling through custom fonts on macOS, per discussion on #278 after it was closed.

Also fixes #283 so I could get it running.


zapan:
I have a problem with this branch in macOS. I cannot type in spaces in "vocabulary meaning" input.

turtlemaster19:
@mzsanford Could you change "fixing" to "fixes" so that it will auto-close #283 and #284 when merged?

For example: "which fixes #284"; "resolves #283"

mzsanford:
@zapan I found the source of that bug and it should be fixed now. Thanks for testing it out.

Issue #285, opened 2020-01-31
macOS: the font size slider at no effect


bmironer:
Changing the font slider seems to have absolutely no effect. Even after relaunching the app

PR #294, opened 2020-02-07
macOS/iPad: Respect font size setting

mzsanford:
This keeps the existing font adjustment for larger screens (250%) but further applies the user setting. Fixes issue #285.


jeff8v7:
The font size setting still has no effect on iOS... Tsurukame iOS version 1.17

turtlemaster19:
@jeff8v7 I can confirm that it **does** work. Will you see if your issue is resolved by downloading the TestFlight link in the README?

89CB397E-CDAA-4D04-A7C2-29E7F427ECFA

jeff8v7:
I just checked in TestFlight on my iPad Pro (13.3.1) , and the font size setting has no effect on the lesson text, or anywhere else I could see.  100% and 250% look exactly the same.

jeff8v7:
100%
29144794-7CFB-4343-900E-4DB6A5167707
C5C9E3AA-4A34-4504-89B5-202BFB308CE6

jeff8v7:
250%
798EF338-49E7-4737-8B52-7172BCE5DD93
E00972B2-184F-419E-B08D-876318EE6397

turtlemaster19:
@jeff8v7 I think that's what this fixed... the latest commits haven't been pushed to TestFlight yet...

jeff8v7:
Thanks! You're right. I just checked the latest TestFlight version 1.18 (681) on my iPad Pro. The font size for the review item adjusts according to the font size setting.

Perhaps I was confused on the intent behind the "font size" setting. On an iPad pro, all the text seems very small in the app, with the exception of the large review item. I was trying to increase the size of all the text in the app (except the review item which is already large). So, naturally I anticipated that the "font size" setting would do just that. 

Now I suspect, the "font size" is a setting for iPhone users, which allows them to change the size of the review item, but not any of the other text - which totally makes sense for iPhone. Unfortunately, however, this may be the opposite of what iPad users would expect/want.

turtlemaster19:
@jeff8v7 Well, the font size of the review item can also feel small on the iPad or Mac as well.

Issue #277, opened 2020-01-22
Reviews are not working


alisonlopez:
Hello! First of all I want to thank you fot this amazing app...It's really useful and user-friendly and it really helped me to get my lessons and reviews in order...but a month ago it stopped working for me. And it's a shame because I think it's the best app in the App Store for wanikani :(

It seems that the app cannot submit my lessons and reviews...If I do some reviews and go to the PC I see the reviews from my app weren't submitted so I have to do them again. Also, if I go to Show remaining section all the cards are blank...so it clearly can connect to wanikani.

My device is an iPhone 7. I tried to uninstall and install again but it didn't solve anything.

Some help please, I'm stuck a level 19 forever because I hardly can go to a pc to do my reviews and other apps are terrible :(

turtlemaster19:
Do you experience this issue on the official TestFlight? By the way, this is covered by #225 and #235, so either this or those will need to be closed.

alisonlopez:
Don't know If it's okay to comment here or in the other issue...

How can I test it?

mzsanford:
@alisonlopez if you have a Mac check out https://github.com/davidsansome/tsurukame/issues/235#issuecomment-577273703 for how to gather logs that can help us find the root cause and get it fixed.

turtlemaster19:
@alisonlopez It might be a good idea to close this issue since there's another issue that this is a duplicate of. You can keep commenting here after it's closed.

alisonlopez:
Yeah, It doesn't work even with the testflight app. 

Sorry I don't have a MAC to get the logs for you :(

davidsansome:
@alisonlopez could you go to Settings in the app, hit "Export local database" at the bottom and email it to me@davidsansome.com?  That should contain a log of the requests that failed.

turtlemaster19:
@davidsansome Dupe of #235

davidsansome:
@alisonlopez sent me their database, it looks like there's a clock skew problem on the client:

```
sqlite> select request_url, response_data from error_log;
https://api.wanikani.com/v2/assignments?unlocked=true&hidden=false&updated_after=2020-01-22T20:53:54.486000Z|{"error":"The updated_after parameter must be in the past","code":422}
https://api.wanikani.com/v2/study_materials?updated_after=2020-01-22T20:53:52.563000Z|{"error":"The updated_after parameter must be in the past","code":422}
https://api.wanikani.com/v2/reviews/|{"error":"Record invalid","code":422}
[...]

PR #295, opened 2020-02-07
Use server time instead of local client time

mzsanford:
Try to avoid clock drift problems by using server time where available. Also avoid sending a client-generated `created_at` when they are sufficiently new.

This addresses the first two suggestions from [this comment](https://github.com/davidsansome/tsurukame/issues/277#issuecomment-577978734) by @davidsansome on #277.


ios/Client.h
```diff
@@ -31,9 +31,11 @@ typedef void (^ApiTokenHandler)(NSError *_Nullable error,
                                 NSString *_Nullable apiToken,
                                 NSString *_Nullable emailAddress);
 typedef void (^AssignmentHandler)(NSError *_Nullable error,
+                                  NSString *_Nullable dataUpdatedAtIOS8601,
```
turtlemaster19: `dataUpdatedAtIOS8601` should probably be `dataUpdatedAtISO8601`. Same with all other uses of it.

Issue #201, opened 2019-11-12
Random Lesson Order


Phrim1:
It would be nice to be able to sort Lessons by random instead of some combination of Radical/Kanji/Vocab.  This is implemented on the web with the option "Lesson Ordering: Shuffled".

zebala:
+1 to this. Recently I've come to hate the order in which WK gives the items, since it groups vocab using the same kanji together. This backfires in reviews as you may fail one but then remember the other ones even though you didn't. IMO it's better to sprinkle them. Also makes reviews and lessons a bit more interesting, and you don't get lots of similar items together.

turtlemaster19:
I was thinking we could do something like the below image, and have the cells be able to move between the two sections. Obviously, shuffling just vocabulary would do nothing, but the idea is that the cells would be able to move between both.

![UI Mockup](https://user-images.githubusercontent.com/46784000/75450389-535d0d00-5934-11ea-96d9-bc0c3e7fd34c.png)

zebala:
Looks pretty nice! Wouldn't the vocab shuffling still change the order of the vocab items? Assuming we don't have any pending radicals or kanji.

turtlemaster19:
@zebala That might actually be a good idea if that were implemented. We would then just completely randomise anything in the shuffle group, and keep our current order outside of it...

Phrim1:
That looks awesome!

zebala:
I wonder if I wanted to do some vocab lessons from the previous level shuffled could be handled. Like in WK "Ascending level then shuffled". Basically if there is vocab spanning multiple levels.

turtlemaster19:
@zebala Good point. I think if we implement shuffling, level should be taken care of first. For example, if prioritize current level is on, that would take precedence over the other shuffling things.

PR #306, opened 2020-02-27
Sort by item type in reviews (Fixes #304)

### LINKED ISSUE #304: Feature Request: Sort by item type in reviews
Just a nice to have; in lessons there's an option to sort which order you want to learn radicals/kanji/vocab.  Would be awesome to have the same ordering functionality in lessons.

zapan:
Sort by item type in reviews. Fix #304

### DIFF
--- a/ios/Resources/Main.storyboard
+++ b/ios/Resources/Main.storyboard
@@ -751,6 +751,23 @@
                                             </subviews>
                                         </tableViewCellContentView>
                                     </tableViewCell>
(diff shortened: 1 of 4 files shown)


turtlemaster19:
@zapan I think if #304 were to be implemented, it should be like lessons where the user can choose any combination of Radical/Kanji/Vocab similar to the UI mockup in #201.

For example, I completely agree with #304 but I would want to use SRS stage order still, and pick a combination.

zapan:
@turtlemaster19 You're right, that would be superb. I'll see what I can do 🤔

turtlemaster19:
@zapan You might as well end up implementing #201 if you do that though.

zapan:
Work in progress to implement @turtlemaster19 suggestions

PR #308, opened 2020-02-28
Guard against ill timed Siri Shortcut invocation

### LINKED ISSUE #307: Bug Report: Lessons from Siri Shortcut with none left
I clicked on the Start Lessons Siri shortcut, but was greeted by a crash screen, since I had 0 remaining.

mzsanford:
Do not attempt to launch lesson or review controller if Siri suggests a user do lessons or reviews when none exist.

Fix #307


ios/AppDelegate.m
```diff
@@ -66,10 +66,26 @@ - (BOOL)application:(UIApplication *)application
     willContinueUserActivityWithType:(NSString *)userActivityType {
   if ([userActivityType isEqual:SiriShortcutHelper.ShortcutTypeReviews]) {
     MainViewController *mainVC = [self findMainViewController];
-    [mainVC performSegueWithIdentifier:@"startReviews" sender:nil];
+    NSArray<TKMAssignment *> *assignments = [_services.localCachingClient getAllAssignments];
```
davidsansome: If you just want to get the number of available lessons and reviews you can use localCachingClient.availableLessonCount and localCachingClient.availableReviewCount.  They're cached already so you don't need to re-calculate them here.

mzsanford: I hadn't noticed those. Updated.

davidsansome: Looks great, thank you!

Issue #338, opened 2020-04-01
Bug: item keeps returning into queue


mark-rossiter:
Observed
======
The kanji item 'hills' (https://www.wanikani.com/radicals/hills) keeps returning into my Review queue every time I finish a review. Even if it is just this one item in the queue, I can go in and finish it and it will say it is 'enlightened' and still go back into the queue. 

Expected
======
Item should be added back into the standard SRS cycle to return at the time it is to be 'burned'

Steps
===
1 Open Tsurekame
2 Complete review
3 App refreshed and there is 1 item popping into the queue
4 That item is https://www.wanikani.com/radicals/hills I can complete it but then it adds back in again

zapan:
Duplicated from #324

zapan:
In the meantime you can use TestFlight, version 1.20 build 693 has been published

turtlemaster19:
@mark-rossiter Please close this issue.

turtlemaster19:
@davidsansome Shouldn't this have been closed by #341?

Issue #324, opened 2020-03-26
Radical does not progress


Rion-Kaneshiro:
I have a radical since yesterday that seems to be stuck on guru one and reappears in reviews all the time although it’s already answered correct.
When I reload after I finished the reviews it appears as a review again.

On the Wanikani website I do not have any open reviews, just in the app.
I deleted and reinstalled the app an there it was again.

turtlemaster19:
@Rion-Kaneshiro Good news! We just released an update to fix this! Please update the app.

davidsansome:
Apple took much longer than usual to approve this new release, it only just went out to the App Store a few hours ago!

Rion-Kaneshiro:
Updated and I can confirm
It’s still happening :/

turtlemaster19:
@Rion-Kaneshiro Really? You're on 1.18 now?

Anyways, it may be useful to now which one it is.

Rion-Kaneshiro:
Yes, latest version from the store.

Would it help you to export the database?

turtlemaster19:
@Rion-Kaneshiro I would be more interested at the moment to know which radical it is.

Rion-Kaneshiro:
https://www.wanikani.com/radicals/hills

turtlemaster19:
@Rion-Kaneshiro Thank God they moved it! I was a part of the thread asking for it to be moved.

In that case, it will be corrected in the next App Store release, which will be out as soon as Apple approves it.

Rion-Kaneshiro:
Cool, thanks :)

davidsansome:
Oops sorry I was mistaken - it looks like Wanikani updated this radical **after** I created the database for 1.18 and 1.19, so the update won't fix the issue.
I'll have to release another update with the new data.

turtlemaster19:
Yep, the official content update thread for it is here:

https://community.wanikani.com/t/march-27-2020-content-updates/42798

jeff8v7:
Same problem. Hills radical is stuck in my review queue. Is there a way to reset a radical's progress to 0 in wanikani? Otherwise, I'm going to know it really really well :-)

turtlemaster19:
@jeff8v7 If it has been burned in WaniKani, it can be resurrected.

Otherwise, I am afraid not. As soon as the next TestFlight build or App Store release comes, neither of which have occurred yet, this issue will go away.

turtlemaster19:
[...]

PR #341, opened 2020-04-12
Stop lessons from higher levels appearing

d108:
This further addresses the problem where the 'hills' radical was always being shown in lessons for some users.

See #324 and #338.


davidsansome:
Thanks for this!

But what about doing it in ReviewItem.m instead?  There's already code that does this in assignmentsReadyForReview - you'd just need to copy it to assignmentsReadyForLesson:
https://github.com/davidsansome/tsurukame/blob/master/ios/ReviewItem.m#L32

d108:
Sure, that makes sense. A reference to a `LocalCachingClient` is not passed into assignmentsReadyForLesson. Would you be fine with it being added for this purpose?

davidsansome:
Yep sounds good to me!

d108:
Should be all set!

davidsansome:
Awesome, thank you!

Issue #257, opened 2020-01-13
Feature Request: Resurrect Burns & view item SRS stage


turtlemaster19:
When in the item view in a search, it should be possible to:
1. See the SRS stage/next review time of the item
2. Resurrect the item if it's burned if the user wants (Should have confirmation, though.)

PR #351, opened 2020-04-30
Create stats section

turtlemaster19:
Partially implements #257 by showing first started date, SRS stage, next review date, and guru date.


turtlemaster19:
Moved to #356

PR #373, opened 2020-07-06
README fixes

UInt2048:
Update the README to include the disclaimer instead, and how to use the shell script introduced with the Apple Watch Complication. This is independent of the other PRs, and can be merged immediately.

P.S. Could you tag 1.21-700 since it has been released to the App Store now?


UInt2048:
@davidsansome Ready for review. This is completely independent of everything else.

P.S. Could you tag 1.21-700 since it has been released to the App Store now?

PR #379, opened 2020-07-23
Change auto-reveal condition (Fix #321)

### LINKED ISSUE #321: Bug/Feature Request? Customise auto reveal setting
So, I have the "Reveal answer automatically" setting set to "Off", yet if I get an answer wrong twice the answer immediately _does_ reveal automatically.  

Presumably this was a design choice yet personally if I had this setting "On" I would like to customise the amount of wrong answers before an auto reveal (e.g. have the aforementioned setting "On" and a new setting "Number of incorrect answers before reveal" set to "4").  

Feels to me like the setting needs expansion basically, since it doesn't work how I'd expect it to at the moment.

UInt2048:
Fix #321. If the setting is off, we should never reveal it automatically, and if it's on, we should always reveal automatically.

### DIFF
--- a/ios/ReviewViewController.swift
+++ b/ios/ReviewViewController.swift
@@ -1135,7 +1135,7 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, TKMSubjectDel
     }
 
     // Otherwise show the correct answer.
-    if !Settings.showAnswerImmediately, firstTimeAnswered {
+    if !Settings.showAnswerImmediately {
       revealAnswerButton.isHidden = false
       UIView.animate(withDuration: animationDuration,
                      animations: {


ios/ReviewViewController.swift
```diff
@@ -1130,8 +1130,10 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, TKMSubjectDel
       return
     }
 
-    // Otherwise show the correct answer.
-    if !Settings.showAnswerImmediately, firstTimeAnswered {
+    // Show the correct answer if it's not the first time.
+    if Settings.showAnswerImmediately, !lastMarkAnswerWasFirstTime {
```
davidsansome, line 1134: If `showAnswerImmediately` is true shouldn't we always show the answer regardless of whether it was the first time or not?
Also why the switch from `firstTimeAnswered` to `lastMarkAnswerWasFirstTime`? I think they're equivalent?

UInt2048: You're right, actually. If the setting is off, we should never reveal it automatically, and if it's on, we should always reveal automatically. Updated the PR to reflect this.

davidsansome: Looks good, thanks!

PR #380, opened 2020-07-28
Set the subject details view to be first responder when needed

### LINKED ISSUE #369: Keyboard focus lost after "Back" navigation
During reviews, the "wrong answer" screen can be dismissed/continued using the `return` key on the keyboard.

However, if you navigate to another screen (say, tapping a kanji to review it) and then go back via swipe or "Back" button, the keyboard focus is lost—pressing `return` no longer does anything. At that point, keyboard navigation is broken until after you have tapped on the "->" button with your finger.

**The screen in question, where `return` works to continue, unless you have gone elsewhere and come back**

62EB4E54-9087-4D47-BFFC-7B9121914B0A

d108:
This treats the case where a SubjectDetailsView can lose its keyboard focus during navigation events in reviews, and the expected keyboard commands do not work.

Fixes #369.

### DIFF
--- a/ios/ReviewViewController.swift
+++ b/ios/ReviewViewController.swift
@@ -390,14 +390,18 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, TKMSubjectDel
     if subjectDetailsView.isHidden {
       answerField.becomeFirstResponder()
       answerField.reloadInputViews()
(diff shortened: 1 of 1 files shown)


davidsansome:
Thanks!

PR #381, opened 2020-08-03
Allow hardware keyboard control during wrong answers

d108:
These changes allow users of hardware keyboards to have continual control of the UI during wrong answers with keyboard commands. This includes the following conditions listed below.

1. Keep the cursor in the text field on AnswerCheckerResult case OtherKanjiReading or ContainsInvalidCharacters, allowing the text to be edited after a shake animation.
2. When an answer is Incorrect, and before display of subject details, allow the backspace, keyboard enter, or numeric keypad enter key to clear the field and allow another entry.
3. When the subject details view is displayed, allow an enter key press to close the view and proceed to a new answer entry.

See #286 for a previous report of (1).


davidsansome:
Awesome, thank you!

d108:
どう致しまして！

PR #384, opened 2020-08-03
Convert TKMOfflineAudioViewController

UInt2048:
This PR converts TKMOfflineAudioViewController to Swift, and switches Light Untar with [Light Swift Untar](https://github.com/Light-Untar/Light-Swift-Untar) so that it will work. (If I didn't, it would complain about `@escaping` closures. 😿)

The dependency change is the only reason the line change count is so high. Sorry about that!


davidsansome:
Sorry I assumed you'd tested this so I committed it without trying it myself.

It doesn't work at all - tapping "Offline Audio" in the Settings view presents a blank audio controller - you need to update the class name (drop the "TKM" prefix) and the Swift module in the storyboard. After fixing that it looks like it tries to download URLs like `file://https:/tsurukame.app/...`. After that I gave up and reverted the PR.

I appreciate you making these changes but you've really gotta test that they actually work first - compile them locally, run them on the simulator and check that they're at least basically functional.

UInt2048:
Oh noes! I thought it worked, too. Evidently, during the monster Swift conversions, I must have forgotten to test this one, but then assumed it worked just fine and copied it straight from there. I will submit a properly tested PR as soon as I can.

UInt2048:
Properly tested and submitted #387. I can't believe I forgot to test that one! 🤦‍♂

PR #385, opened 2020-08-03
Formatting changes

UInt2048:
1. Moves SwiftFormat to SwiftPM, and introduces an up to next minor (e.g. `~> 0.45.0`) requirement to avoid surprises like at 0.45.0. If you would prefer, this could be raised to next major.
2. Removes `clang-format-all.sh` by inserting its contents into the Format Code build phase.
3. Moves `NSDate+TimeAgo` to the Extensions folder, and moves `proto/Wanikani+Convenience` to `Extensions/ProtobufExtensions`.
4. Prohibits Swift Format from formatting inside the `proto` folder.
5. Runs `go mod tidy` which probably should have been done with the dependency updates, but I forgot to do it.

Again, the huge line count is due to moving it away from CocoaPods.


UInt2048:
Ready for review now!

UInt2048:
@davidsansome There's no merge conflict problem now. Ready for review!

Issue #371, opened 2020-07-05
Bug: Broken HTML in explanation of 妖 in lesson


Mercieral:
Im seeing some weird HTML sneaking into this lesson page.

![7D3C6E15-2DE8-47FC-B9E9-CA2EFBC8B779](https://user-images.githubusercontent.com/10813646/86523657-eeab4500-be34-11ea-9424-dab1a82ad17e.png)

turtlemaster19:
@Mercieral Issue on WK's side. See https://community.wanikani.com/t/妖-invalid-tags/44741

PR #389, opened 2020-08-05
Update Protobuf

UInt2048:
This PR updates Protobuf to 4.0.0-rc1 and syntax to Proto3.


proto/wanikani.proto
```diff
@@ -191,14 +195,14 @@ message SubjectOverrides {
...
   enum Format {
-    RADICAL = 1;
-    KANJI = 2;
-    JAPANESE = 3;
-    READING = 4;
-    VOCABULARY = 5;
-    ITALIC = 6;
-    BOLD = 7;
-    LINK = 8;
+    RADICAL = 0;
```
davidsansome, line 198: Changing these field numbers is probably fine here as long as we recreate the data.bin at the same time, but it still makes me nervous. Can we keep them as they were before but add a new `UNKNOWN = 0`?

UInt2048: Absolutely! However, I would still appreciate a data.bin update (which would immediately fix #371)

UInt2048:
Thank you! That's exactly what needed to be fixed.

PR #392, opened 2020-08-09
Add edit meaning/reading notes & show all buttons (Fixes #27, #172, #361)

### LINKED ISSUE #27: Allow users to make notes on meanings
I would like to be able to see the notes I have on meanings and create new ones.

### LINKED ISSUE #172: Emulating Wanikani behavior on wrong answer
When you get a review wrong for a meaning/reading, Wanikani’s webapp only shows you the correct answer for the half you got wrong.

This is good because you don’t see e.g the meaning if you got the reading wrong.

This is an issue with typos — sometimes i might make a typo, mark it as correct, and see the other field before having completed that half of the review. 

Ideally, Tsurukame would only show you the part you got wrong, and let you tap to view the full page.

### LINKED ISSUE #361: Feature request: improve the user note feature
First of all, I'd like to say that I like the Tsurukame app very much, I use it all the time.

Recently, Tsurukame started to show the user note field in the kanji/vocabulary information screen.  I must say, it was a life saver for me.  I use user notes very heavily, and before that I had to switch to the WaniKani app in the mobile Chrome constantly, which was very time consuming.  So thank you for introducing this feature!

I'd like to request a couple improvements:

1. There are two different note fields in WaniKani, the "Meaning Note" field and the "Reading Note" field.  However, Tsurukame only shows one field instead of two.  What would happen if both these fields are non-empty in WaniKani? 

In any case, please **show these two note fields separately** in Tsurukame as well.  I'd say show them below the corresponding "Meaning Mnemonic" and "Reading Mnemonic" fields, just like they are shown in WaniKani.

2. Please enable **adding and editing** the "Meaning Note" and "Reading Note" fields via the Tsurukame app interface.  This way, this feature will be complete.

Thank you.

UInt2048:
Fixes #172 by adding a button to show meaning & reading, as well as correcting some bugs created in #383.

Also adds edit meaning/reading notes buttons, which fixes #27 and closes #361 (a duplicate), and corrects reading note bug (migrated from #405).

### DIFF
--- a/ios/SettingsViewController.swift
+++ b/ios/SettingsViewController.swift
@@ -297,7 +297,7 @@ class SettingsViewController: UITableViewController {
   @objc private func prioritizeCurrentLevelChanged(_ switchView: UISwitch) {
     Settings.prioritizeCurrentLevel = switchView.isOn
   }
-  
+
   @objc private func showStatsSectionChanged(_ switchView: UISwitch) {
     Settings.showStatsSection = switchView.isOn
   }
--- a/ios/SubjectDetailsView.swift
+++ b/ios/SubjectDetailsView.swift
@@ -135,6 +135,11 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
   private var tableModel: TKMTableModel?
   private var lastSubjectChipTapped: SubjectChip?
 
+  private var subject: TKMSubject!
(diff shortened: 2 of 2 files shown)


UInt2048:
Migrated to #458

PR #382, opened 2020-08-03
Settings conversion

UInt2048:
Converts `Settings` to Swift, and makes necessary adjustments in the imports of other files for the conversion.

PR #395, opened 2020-08-10
Convert Settings VC to Swift & adds functionality to exact match setting (Fixes #86)

### LINKED ISSUE #86: Close but no cigar & disable partial answer match

UInt2048:
Convert Settings View Controller to Swift.

Also fixes #86 by adding functionality to stub exact match setting prematurely introduced in #382.


UInt2048:
Ready for review - I have confirmed that it works.

UInt2048:
View controller conversion superseded by 76ce5d2bb1512bcb74c3fd09ef01f8a8b64c960e

PR #399, opened 2020-08-16
Convert TKMFontLoader to Swift

UInt2048:
Swift conversion of `TKMFontLoader`


UInt2048:
Ready for review - confirmed that it works.

PR #402, opened 2020-08-23
Support wrong answer counts greater than one

d108:
## Summary

These revisions make changes to wrong answer handling for the following conditions.

- Incorrect answers beyond the first one for meanings and readings now get recorded and submitted to the Wanikani API.
- If, after submitting an incorrect answer and skipping the subject, the answer counts in the review queue persist instead of being discarded.
- When a subject reappears, where previously a reading or meaning has been entered correctly, a prompt for the correct part will not appear.

## Motivation

When studying with grouped meanings and readings (back-to-back), entering an incorrect meaning, viewing its details, and then skipping the subject, it reappears at the end of the review queue. However, wrong answers for the subject not recorded resulted in SRS progress advancing mistakenly. It could reach the "burned" stage even after multiple incorrect guesses during the review.


d108:
@davidsansome The `wanikani.proto` source does not appear to be in the project. Therefore, I have made changes directly to the generated files. Feel free to overwrite them with newly created ones if this is an acceptable PR.

UInt2048:
@d108 No, `wanikani.proto` is in the `proto` subdirectory of the project root. You can't have this PR be merged without editing that.

d108:
@Uint2048, thanks for pointing me to the file. I failed to look outside of the Xcode project! I will update the PR after taking a look at it and also consider offering the new behavior as an option.

UInt2048:
@d108 I understand your perspective, but from mine, it can be a feature that Tsurukame will do this, reducing the penalty for getting something wrong many times. Hence the request that it becomes an option!

UInt2048:
@d108 Added a PR so you don't have to do the work of installing Protobuf!

d108:
@UInt2048 Thanks for adding your changes and making this PR ready for review. I would have added a setting when I had some time because there is a good reason for it, but you saved me the effort.

UInt2048:
@d108 No problem, please see my feedback.

d108:
@UInt2048 If you are referring to your comment about reaching "burned" after multiple wrong guesses not happening, I did not immediately notice it here, perhaps due to it not being added at the same time as your original reply.

The condition is possible when skipping subjects and `activeTask.reset()` is invoked. By removing that call, the task state persists, so that wrong answers entered before a skip get recorded to WK when the subject is complete.

ios/ReviewViewController.swift
```diff
@@ -1054,7 +1054,6 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, TKMSubjectDel
     if result == .AskAgainLater {
       // Take the task out of the queue so it comes back later.
       activeQueue.remove(at: activeTaskIndex)
-      activeTask.reset()
```
UInt2048, line 1057: That's a good idea to remove the `activeTask.reset()`, but if we do, we should make sure that this item comes back in the queue before the session ends (e.g. by the wrap-up button)

d108: @UInt2048 If I understand your concern correctly, it is that the review queue handling will change due to not calling reset on the active task.

Showing that this PR does not affect that functioning is possible by selecting "wrap up" during reviews while changing various related settings. Are you finding different behavior?

UInt2048: @d108 My point is that since the skipped item is removed from the queue, it could still contain partial progress after a wrap-up (since it's not in the queue now, wrap-up won't ask about it)

The app used to solve this problem by resetting that partial progress on this line...

d108: @UInt2048 If skipped subjects are not re-queued that will defeat the purpose of having the feature.

The next line ensures this does not happen.

https://github.com/davidsansome/tsurukame/blob/fd9da51a3e5f758dd5f72637b255a5298788787e/ios/ReviewViewController.swift#L1057

The change in behavior is that a previously correct meaning or reading does not require input at the end of the review queue or during a wrap-up.

ios/ReviewViewController.swift
```diff
@@ -1078,13 +1077,25 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, TKMSubjectDel
         activeTask.answer.meaningWrong = !correct
       }
       activeTask.answeredMeaning = correct
+
+      // A false answeredMeaning indicates an incorrect answer.
```
davidsansome, line 1081: You could just use `if !correct` here then you wouldn't need the comment

ios/ReviewViewController.swift
```diff
@@ -1078,13 +1077,25 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, TKMSubjectDel
         activeTask.answer.meaningWrong = !correct
```
davidsansome, line 1077: OverrideAnswerCorrect should also reset the wrongCount to 0

ios/Settings.swift
```diff
@@ -96,6 +96,7 @@ import Foundation
   @Setting(false, #keyPath(showAllReadings)) static var showAllReadings: Bool
   @Setting(false, #keyPath(autoSwitchKeyboard)) static var autoSwitchKeyboard: Bool
   @Setting(false, #keyPath(allowSkippingReviews)) static var allowSkippingReviews: Bool
+  @Setting(false, #keyPath(minimizeReviewPenalty)) static var minimizeReviewPenalty: Bool
```
davidsansome, line 99: Can we default this to true so it keeps the existing behaviour?

ios/ReviewViewController.swift
davidsansome: "Ask Again Later" is a cheat, it's supposed to forget your wrong answer and add the question back again at the end of the queue. It should reset the wrong counts as well - can you add this `activeTask.reset()` line back in?

UInt2048: @d108 If you wanted to change the behavior to remember prior progress before the removed answer - that should be a separate PR.

UInt2048:
@d108 See David Sansome's review

ios/ReviewViewController.swift
UInt2048: There's a fundamental difference between `activeQueue` and `reviewQueue`... the changes you would need to make to keep all previous progress, without forgetting to ask again in the same session, would need to be a separate PR.

UInt2048:
@d108 Created PR for you to review - should fix these issues

ios/ReviewViewController.swift
d108: There are a couple of different lines of reasoning running into each other here. I will restore it to just supporting wrong answer counting, but that was not my motivation behind the changes. Hopefully, the latest commit will address all of the concerns brought up while creating a foundation to implement the intended purpose in a new PR.

PR #413, opened 2020-10-02
Tsurukame Widget Extension

### LINKED ISSUE #376: Tsurukame Widget
Similar to the Watch complication introduced in #289, I propose adding an iOS widget showing either the lesson & review counts, or the level, depending on the setting chosen in the app.

For iOS 14.0+ and macOS 11.0+, we should use the new WidgetKit as documented [here](https://developer.apple.com/documentation/widgetkit)

UInt2048:
This pull request fixes #376 by adding a WidgetKit widget, which supports by necessity only macOS 11+ and iOS 14+.

Adding the prior Today View iOS widget is beyond the scope of this PR, since it's already large enough.

As previously discussed in #270, adding the App Groups capability was in fact required to implement this. However, instead of moving the existing caching client, this PR proposes to simply save/load the data to/from a file.

### DIFF
--- a/ios/AppGroup.swift
+++ b/ios/AppGroup.swift
@@ -1,0 +1,210 @@
+// Copyright 2021 David Sansome
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


UInt2048:
@davidsansome This PR is finally ready for review. About half of this is just adding to `project.pbxproj` or required assets stubs, so it's not actually as bad as you might think.

Please let me know if there's anything I can do to help!

UInt2048:
This PR will be ready immediately upon merging #507

UInt2048:
Closing draft PR, will reopen if ready.

Issue #313, opened 2020-03-05
Long words get cutoff (with ellipsis) during review


jeff8v7:
On iOS (iPhone X), review words longer than 3 letters get cut off with an ellipsis, which makes for a fun guessing game.

Update: Still seeing the ellipsis for 4 character words if font size is 250% (on iPhone X, Tsurukame Version 1.20 (693).  But any other font size is fine (225% is fine).   I sent a screenshot from testflight.

Seems unlikely anyone will every need 250% font size, but still....

 
![E8ACDF0F-7E9B-447E-92E6-D4E4D61A9C91](https://user-images.githubusercontent.com/39231776/75976453-cd395b80-5f1d-11ea-9bf0-6da8ddc0d078.jpeg)

turtlemaster19:
@jeff8v7 It might help if you decrease the font size for now, 250% seems a bit ridiculous for an iPhone X.

jeff8v7:
Actually, this was with font size 100%, because the font size wasn't working when I took this screen shot.  Today's update 1.18 (681) changed the behavior, so that font size does make this bigger (but strangely doesn't affect any other fonts in the app)

turtlemaster19:
@jeff8v7 That's intended behavior. Any other behavior should receive its own feature request.

turtlemaster19:
@jeff8v7 Are you still having this problem?

jeff8v7:
Not sure. I don’t usually review in iPhone, but I’ll try some and see if I can get this issue again.

jeff8v7:
Still seeing the ellipsis for 4 character words if font size is 250% (on iPhone X, Tsurukame Version 1.20 (693).  But any other font size is fine (225% is fine).   I sent a screenshot from testflight.

Seems unlikely anyone will every need 250% font size, but still....

turtlemaster19:
@jeff8v7 Would you mind updating the first post of the issue with this?

turtlemaster19:
@jeff8v7 It is an important issue, especially for `お誕生日おめでとう` which fills the entire screen at 100%.

PR #414, opened 2020-10-03
Make review label number of lines 2

abdulajet:
Quick fix for issue #313. Made number of lines 2 instead of one. Maybe it should be 0? 

I used Xcode 12.0.1 so that might have caused the extra storyboard changes

Screenshots on 11 pro with 125% font size
| Before        | After           |
| ----------- |------------|
| Simulator Screen Shot - iPhone 11 Pro - 2020-10-03 at 17 25 17 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-03 at 17 24 45 |

### DIFF
--- a/ios/Resources/Main.storyboard
+++ b/ios/Resources/Main.storyboard
@@ -350,7 +350,7 @@
                                     <constraint firstAttribute="trailing" secondItem="kc6-Lm-8sK" secondAttribute="trailing" constant="8" id="hnc-JP-Evv"/>
                                 </constraints>
                             </view>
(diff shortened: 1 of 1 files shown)


UInt2048:
@abdulajet For reference, the relevant change is on line 354.

Would it work with zero? If so, that might be useful for larger font size (e.g. 250%)

abdulajet:
> @abdulajet For reference, the relevant change is on line 354.

want me to commit only this line?

I can do 0.

UInt2048:
@abdulajet Can you? Try that and see if it works, since if it would need three lines in a large font, that would help.

abdulajet:
When setting it to 0 you run the risk of the bottom sheet not coming up when you get the answer wrong. Is お誕生日おめでとう the longest string? I don't think so

Simulator Screen Shot - iPhone 11 Pro - 2020-10-03 at 18 40 36

PR #438, opened 2021-03-12
Implement proposed fix for #436

### LINKED ISSUE #436: Attempting to log in with a password containing an & fails, because the & breaks the arguments in the web request.
When attempting to log in with a password containing the character &, the login request fails.
This seems to be happening because the data in the HTTP body is not encoded.

Cause of the issue appears here:
https://github.com/davidsansome/tsurukame/blob/0060ec8b352cbc3216c2b69588c0171826d4d7d6/ios/WaniKaniAPI/WebClient.swift#L101

I'm unable to change and test this myself, as I don't have a working dev env set up on a mac.
I did do some testing and think this might, potentially, solve it.
Changing:
```
      // Build the login request.
      let queryItems = [
        URLQueryItem(name: "user[login]", value: username),
        URLQueryItem(name: "user[password]", value: password),
        URLQueryItem(name: "user[remember_me]", value: "0"),
        URLQueryItem(name: "authenticity_token", value: csrfToken),
        URLQueryItem(name: "utf8", value: "✓"),
      ]

      var req = URLRequest(url: kLoginUrl)
      req.httpShouldHandleCookies = true
      try req.setFormBody(method: "POST", queryItems: queryItems)
      return request(req, session: session)
```

into

```
        var req = URLRequest(url: kLoginUrl)
        req.httpMethod = "POST"
        req.setValue("application/x-www-form-urlencoded;charset=UTF-8", forHTTPHeaderField: "Content-Type")

        var urlComponents = URLComponents()
        urlComponents.queryItems = [
          URLQueryItem(name: "user[login]", value: username),
          URLQueryItem(name: "user[password]", value: password),
          URLQueryItem(name: "user[remember_me]", value: "0"),
          URLQueryItem(name: "authenticity_token", value: csrfToken),
          URLQueryItem(name: "utf8", value: "✓"),
        ]
        request.httpBody = urlComponents.percentEncodedQuery?.data(using: .utf8)
        return request(req, session: session)
```

Again, I'm not 100% sure this fixes it because I can't test this locally myself, but I'm hoping someone who does have a working setup can build on this suggestion.

UInt2048:
Fixes #436 as suggested in the issue's discussion

### DIFF
--- a/ios/AnswerTextField.swift
+++ b/ios/AnswerTextField.swift
@@ -1,4 +1,4 @@
-// Copyright 2020 David Sansome
+// Copyright 2021 David Sansome
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/ios/ComplicationExtension/ComplicationController.swift
+++ b/ios/ComplicationExtension/ComplicationController.swift
@@ -1,4 +1,4 @@
-// Copyright 2020 David Sansome
+// Copyright 2021 David Sansome
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/ios/ComplicationExtension/DataManager.swift
+++ b/ios/ComplicationExtension/DataManager.swift
@@ -1,4 +1,4 @@
-// Copyright 2020 David Sansome
+// Copyright 2021 David Sansome
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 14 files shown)


UInt2048:
@davidsansome Please comment. I believe this is ready to merge.

davidsansome:
Unfortunately this PR would undo the fix put in for + characters in d17ea37a16559b77159764d6f98b79f910624aec. I've committed d6cef7f3ad94a06d6bd2266cf12f942b61c9f516 instead which fixes both!

PR #439, opened 2021-03-12
Fallthrough in minimum time to guru to fix wonko review times

### LINKED ISSUE #434: Review times seem wrong
This might be an issue on the wanikani side but i have a kanji on a different SRS level that says it will be completed at the same time as the kanji on the above level. image

UInt2048:
Fixes #434 by correcting `minimumTimeUntilGuru(itemLevel:)`


davidsansome:
Oops! Thanks for fixing this!

PR #449, opened 2021-04-06
Add option to hide stats section

### LINKED ISSUE #445: Hirobaba changed to Katakana
is:issue is:open many hiragana words have become katakana. The app also forces me to type those words in katakana. The app thinks these answers are correct. For example: West is originally せい。 but the app forces the user to type セイ. The system accepts this as a correct answer.
E472E31B-508A-4278-A90E-0956C94FA0BD

### LINKED ISSUE #448: Tsurukame problem
Is there a way to turn the Wanikani stats off. And also is this a bug? Why doesn’t it show the member kanji as visually similar when looking at the buy kanji. I’ve looked at other kanji’s and it is also the same with them. Not all. But most 
AAEFD607-AEC5-44F3-A63F-AF0C47F8ED2E

UInt2048:
Fixes #448 by adding a new option to hide the stats section. The stats section will be on by default, to keep it the same way it is right now.

This also fixes #445 by reversing the default for `useKatakanaForOnyomi` accidentally changed in #382

### DIFF
--- a/ios/SettingsViewController.swift
+++ b/ios/SettingsViewController.swift
@@ -87,6 +87,12 @@ class SettingsViewController: UITableViewController {
                                 accessoryType: .disclosureIndicator,
                                 target: self,
                                 action: #selector(didTapLessonBatchSize(_:))))
+    model.add(TKMSwitchModelItem(style: .subtitle,
+                                 title: "Show stats section",
+                                 subtitle: "Show view with level, SRS stage, and more",
+                                 on: Settings.showStatsSection,
+                                 target: self,
+                                 action: #selector(showStatsSectionChanged(_:))))
 
     model.addSection("Reviews")
     model.add(TKMBasicModelItem(style: .value1,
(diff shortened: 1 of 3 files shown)


Haris-senpai:
@UInt2048 this is my first time ever on GitHub. How do I add this or does this roll out with an update for the app?

UInt2048:
@Haris-senpai This will later be added in a future update

ios/SettingsViewController.swift
```diff
@@ -87,6 +87,12 @@ class SettingsViewController: UITableViewController {
                                 accessoryType: .disclosureIndicator,
                                 target: self,
                                 action: #selector(didTapLessonBatchSize(_:))))
+    model.add(TKMSwitchModelItem(style: .subtitle,
+                                 title: "Show stats section",
+                                 subtitle: "Show view with level, SRS stage, and more",
```
davidsansome, line 92: I feel like "view" is a bit of a technical term. Can you change "view" to "section"?

davidsansome (approved):
Thanks for the quick fix, and also for fixing #445!

ios/SettingsViewController.swift
UInt2048: Done

UInt2048:
@davidsansome I changed the word "view" to "section" as you requested. This should be good to go for merging! 👍🏻

PR #458, opened 2021-04-13
Anki mode (Fixes #347)

### LINKED ISSUE #86: Close but no cigar & disable partial answer match

### LINKED ISSUE #347: Feature Request: Anki mode
A special mode that works like Anki. Tapping on the screen reveals the answer (reading or meaning depending on the question but not both - to avoid spoiling the following questions). On the answer screen there would also be "Correct" / "Incorrect" buttons that marks the answer (or even maybe Tinder-style swipe left / right). 

This would basically be equivalent of submitting an incorrect answer, then hitting "My Answer Was Correct", but without the reveal of the full answer card and with an optimized UI intended for this interaction. This would definitely speed up reviews as it is a burden to type the answers every time.

UInt2048:
Fixes #347 by implementing Anki mode

### DIFF
--- a/ios/ReviewViewController.swift
+++ b/ios/ReviewViewController.swift
@@ -578,12 +578,18 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, SubjectDelega
         promptGradient = TKMStyle.meaningGradient as! [CGColor]
         promptTextColor = kMeaningTextColor
         taskTypePlaceholder = "Your Response"
+        if Settings.ankiMode {
+          taskTypePlaceholder = "Show answer"
+        }
       case .reading:
         kanaInput.enabled = true
         taskTypePrompt = "Reading"
         promptGradient = TKMStyle.readingGradient as! [CGColor]
         promptTextColor = kReadingTextColor
         taskTypePlaceholder = "答え"
+        if Settings.ankiMode {
+          taskTypePlaceholder = "答えを見せる"
+        }
       }
 
       // Choose a random font.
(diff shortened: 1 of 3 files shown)


UInt2048:
Here's a video demoing all the new features: https://i.imgur.com/rwXG6EV.mp4

UInt2048:
@davidsansome Ready for review

UInt2048:
@davidsansome Is there anything else you need from me before you merge this?

I'd be happy to help with whatever it is...

UInt2048:
The remaining former contents of this pull request has been split into multiple PRs.

* This PR fixes Anki mode.
* Editing notes has been migrated to #479 
* Pausing on partially correct answers has been migrated to #482 
* Exact match on burns only isn't being requested, because some other PRs really need to be merged first.

UInt2048:
The remaining former contents of this pull request has been split into multiple PRs.

* This PR fixes Anki mode.
* Editing notes has been migrated to #479 
* Pausing on partially correct answers has been migrated to #482 
* Exact match on burns only isn't being requested, because some other PRs really need to be merged first.

@davidsansome This PR and the PRs just created today are ready for review.

ios/ReviewViewController.swift
```diff
@@ -1011,6 +1017,13 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, SubjectDelega
   }
 
   func submit() {
+    if Settings.ankiMode {
+      // Mark the answer incoorect to show the details. This can still be overriden.
```
davidsansome, line 1021: nit: s/incoorect/incorrect/

davidsansome (approved):
This is awesome! Thanks for splitting it up into separate PRs!

ios/ReviewViewController.swift
UInt2048: Thanks! I'm not sure how that typo escaped me.

PR #465, opened 2021-05-25
Sequential audio playback

realericmc:
added playing audio readings sequentially, and a setting to pick if they should be random


realericmc:
not sure about this bundle test failure.  I didn't think I included anything that would affect this.  That said, it has been a while since I messed with GitHub so any comments on this would be appreciated.

UInt2048:
@realericmc I got tests working with a one-line commit: 368c7b766848d26158a242ea278785475b519d53

Please accept realericmc/tsurukame#1

UInt2048:
@realericmc You just got unlucky, this test failure would have occurred to whoever submitted the next PR/commit.

I didn't test anything else in your PR though, apart from the one-line CI fix.

realericmc:
good to know I wasn't going crazy at least.  thanks for running that down!

ios/SettingsViewController.swift
```diff
@@ -210,6 +210,12 @@ class SettingsViewController: UITableViewController {
                                  on: Settings.playAudioAutomatically,
                                  target: self,
                                  action: #selector(playAudioAutomaticallySwitchChanged(_:))))
+    model.add(TKMSwitchModelItem(style: .subtitle,
```
davidsansome, line 213: I don't think it makes sense to expose this as a setting. The "order" of the audio IDs in the proto is arbitrary so I don't think it's something that users should have to decide about. I suspect this PR is more about avoiding having the same speaker play twice in a row?

How about removing the random option entirely, and making sequential playback the default?

ios/WaniKaniAPI/Sources/WaniKaniAPI/ProtobufExtensions.swift
```diff
@@ -249,6 +249,22 @@ public extension TKMSubject {
     let idx = arc4random_uniform(UInt32(vocabulary.audioIds.count))
     return Int(vocabulary.audioIds[Int(idx)])
   }
+
+  func nextAudioID() -> Int {
+    if !hasVocabulary || vocabulary.audioIds.count < 1 {
+      return 0
+    }
+    enum audioIdx {
+      static var idx: Int = 0
```
davidsansome, line 258: I don't really like the static variable here. I would remove this function (as well as `randomAudioID`) entirely, and then add a `private var lastPlayedAudioIndex` to the `Audio` class, and do this logic in `Audio`'s `play` method.

davidsansome (changes requested):

UInt2048:
I think I implemented the requested changes. I have a direct PR to this repo in #466, as well as a PR to this PR in realericmc#2

ios/WaniKaniAPI/Sources/WaniKaniAPI/ProtobufExtensions.swift
realericmc: agree, I like your way better also

UInt2048:
@davidsansome The requested changes have been implemented.

PR #468, opened 2021-06-03
Fix Settings implementation of setArchiveData for iOS 10

UInt2048:
The original lines in `setArchiveData`, from the original Swift conversion, inadvertently stores settings in a format that `getArchiveData` can't read for iOS 10 users. This PR corrects this issue.


davidsansome:
Thanks for fixing this!
I don't seem to be able to install an iOS 10 simulator through xcode any more though so I can't test it :( There were only 3 iOS 10 users over the last 30 days though so we can probably drop support for it at some point

UInt2048:
Ah, that's unfortunate. I still have it even in Xcode 12, but that's because I installed it in a prior version of Xcode.

I also have a real iPhone 5 (my legacy device) that I do reviews on (using my fork) which is where I discovered the issue. ;)

PR #254, opened 2020-01-07
Show next current level review (Fixes #20)

turtlemaster19:
In order to fix #20, this commit shows the next current level review.

PR #470, opened 2021-06-20
Ignore passed items in current level

UInt2048:
To follow with the original reasoning in #254, this PR will exclude assignments from being included that have already passed, since they are no longer relevant for level-up purposes.


davidsansome:
Sorry I must've missed this one!

PR #472, opened 2021-06-22
Longest relative wait (Fixes #355)

### LINKED ISSUE #355: Feature Request: "Longest Relative Wait" Order
Your lowest-level items are the ones that need to be reviewed most frequently in order to remember them, so weighing all items’ time in the review queue equally will make it more likely to answer wrong on low-level items compared to high-level ones.

Instead, for example, say you order by ratio, and you have an Apprentice 4 item (2 day wait to review it) that’s been sitting in your queue for a week. That item’s ratio is 3.5 (7/2), as that’s how many times longer you’ve waited to review it than recommended.

Then, say you also have a Guru 1 item (1 week wait) that’s been in queue for a month, so it has a ratio of 4 (28/7). That item then, since it has a higher ratio, has spent a longer relative time in the queue than the Apprentice item, and thus would be reviewed first under ratio sorting rules.

UInt2048:
Fixes #355 by adding longest relative wait order.

### DIFF
--- a/ios/Resources/Settings.storyboard
+++ b/ios/Resources/Settings.storyboard
@@ -169,6 +169,23 @@
                                             </subviews>
                                         </tableViewCellContentView>
                                     </tableViewCell>
+                                    <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" preservesSuperviewLayoutMargins="YES" selectionStyle="blue" indentationWidth="10" textLabel="xUD-33-BDY" style="IBUITableViewCellStyleDefault" id="Cme-CY-xiB">
+                                        <rect key="frame" x="0.0" y="336" width="375" height="44"/>
+                                        <autoresizingMask key="autoresizingMask"/>
+                                        <tableViewCellContentView key="contentView" opaque="NO" clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="center" preservesSuperviewLayoutMargins="YES" insetsLayoutMarginsFromSafeArea="NO" tableViewCell="Cme-CY-xiB" id="kLg-JQ-CMv">
+                                            <rect key="frame" x="0.0" y="0.0" width="375" height="44"/>
+                                            <autoresizingMask key="autoresizingMask"/>
+                                            <subviews>
+                                                <label opaque="NO" multipleTouchEnabled="YES" contentMode="left" insetsLayoutMarginsFromSafeArea="NO" text="Longest relative wait" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" id="xUD-33-BDY">
+                                                    <rect key="frame" x="16" y="0.0" width="343" height="44"/>
+                                                    <autoresizingMask key="autoresizingMask"/>
+                                                    <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                                    <nil key="textColor"/>
+                                                    <nil key="highlightedColor"/>
+                                                </label>
+                                            </subviews>
+                                        </tableViewCellContentView>
+                                    </tableViewCell>
                                 </cells>
                             </tableViewSection>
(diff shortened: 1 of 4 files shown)


UInt2048:
@davidsansome Could you take a look at this one?

ios/WaniKaniAPI/Sources/WaniKaniAPI/ProtobufExtensions.swift
```diff
@@ -131,23 +131,50 @@ public enum SRSStage: Int, CustomStringConvertible, Comparable, Strideable {
     }
   }
 
+  public static func duration(_ stage: SRSStage, _ accelerated: Bool) -> TimeInterval {
+    TimeInterval(hours(stage, accelerated) * 60 * 60)
+  }
+
+  private static func hours(_ stage: SRSStage, _ accelerated: Bool) -> Int {
```
davidsansome, line 138: It would be nicer to make this a non-static member function, make it do the `accelerated` (`itemLevel <= 2`) calculation itself, and have it return a TimeInterval. Both uses of this function do all those things anyway.

Something like:

```
public func duration(atItemLevel: Int) -> TimeInterval
```

ios/Resources/Settings.storyboard
```diff
@@ -169,6 +169,23 @@
...
+                                        <rect key="frame" x="0.0" y="336" width="375" height="44"/>
+                                        <autoresizingMask key="autoresizingMask"/>
+                                        <tableViewCellContentView key="contentView" opaque="NO" clipsSubviews="YES" multipleTouchEnabled="YES" contentMode="center" preservesSuperviewLayoutMargins="YES" insetsLayoutMarginsFromSafeArea="NO" tableViewCell="Cme-CY-xiB" id="kLg-JQ-CMv">
+                                            <rect key="frame" x="0.0" y="0.0" width="375" height="44"/>
+                                            <autoresizingMask key="autoresizingMask"/>
+                                            <subviews>
+                                                <label opaque="NO" multipleTouchEnabled="YES" contentMode="left" insetsLayoutMarginsFromSafeArea="NO" text="Longest relative wait" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" id="xUD-33-BDY">
```
davidsansome, line 179: I feel like "Longest relative wait" will need some more explanation here. It took 3 paragraphs in #355 to explain what it does.

Can you add a subtitle to this item with a 1-sentence description?
Maybe something like "Sort by amount of time since each item became available relative to its SRS level timing" ... I'm not sure that's very helpful though - maybe you have a better explanation?

davidsansome (approved):

ios/WaniKaniAPI/Sources/WaniKaniAPI/ProtobufExtensions.swift
UInt2048: Good idea. This has been implemented.

ios/Resources/Settings.storyboard
UInt2048: Done. You might need to tweak it a little since I'm not the best at storyboards :wink:

ios/WaniKaniAPI/Sources/WaniKaniAPI/ProtobufExtensions.swift
```diff
@@ -131,27 +131,39 @@ public enum SRSStage: Int, CustomStringConvertible, Comparable, Strideable {
...
+    case .enlightened:
+      return 10364400
+    case .unlocking, .burned:
+      return 0
+    }
+  }
+
+  public func minimumTimeUntilGuru(itemLevel: Int) -> TimeInterval {
+    var time: TimeInterval = 0, stage = self
+    while stage.category == .apprentice {
```
davidsansome, line 162: That's clever!

Issue #112, opened 2019-04-01
A bit more info


tedkulp:
This is by far my favorite Wanikani app.  But, sometimes I have to pop open AlliCrab just for two bit of information and am wondering if it's possible to add them.

1. Upcoming reviews in a list.  The graph is hard to read when you're really just looking for a count of what is coming up today.  Maybe the graph getting more info would be fine, but it's also nice to see what's coming up tomorrow.
1. Current counts of the different statuses.  Particularly how many are in Apprentice, since I generally use that to decide if I'm going to start some new lessons or not (I try to keep my Apprentice count below 100 so I don't accrue too many reviews).

Keep up the great work.  I know it's tough adding new features especially if they require a whole new screen.  But it's worth asking...  Thanks!

turtlemaster19:
Your second point has been partially added, but Tsurukame's "Novice" a.k.a. "Apprentice 1" isn't listed anywhere in the count. The first point I 100% agree should be added.

turtlemaster19:
@davidsansome, here is a possible picture of what it could look like. It would be very useful to see this number IMHO.
![7FB0256B-9C5A-4729-8712-F9DFE0387560](https://user-images.githubusercontent.com/46784000/69887388-42a4e880-12ac-11ea-831e-6db426cc6bbf.jpeg)

turtlemaster19:
The second point has now been implemented.

turtlemaster19:
A few notes a year later:

1. This is the request for the review forecast timeline details. This is definitely a good idea, but please don't create a new issue for it.
2. This is pretty much implemented. What I would like to see, though, is the total Apprentice count, as well as Apprentice I, II, III, IV counts.

Mercieral:
+1 for the second request here of showing the sub-stage breakdowns of apprentice and guru.

Some potential options I'm thinking of...

- Making the apprentice and guru list items at the bottom of the dashboard clickable (with some form of UI to indicate that they are clickable of course) and expand to show sub-srs level counts.
- Showing the sub counts inline on the apprentice and guru list items. (keep in mind narrow phone screens...)
- Setting to change the count between total vs sub-level counts.

PR #474, opened 2021-06-30
Upcoming reviews list view (Fixes #274)

### LINKED ISSUE #274: Breakdown the review count by SRS level
Since level ~40 I’m having trouble keeping up with reviews. So I try to at least do my Apprentice and Guru reviews, and pick at the higher levels when I can. Sorting reviews by SRS level is perfect for that. We have a total tally of the cards by SRS level and now ‘next level up review’ gives me a idea of when I have Apprentice lessons coming up.

But I have no idea of how many Guru, Master or Enlightened review I have queued up and if it’s getting better or worse. 

The details could be added to the All level list such as :

ALL LEVELS          REVIEWS / TOTALS
Apprentice             123 / 136
Guru                       834 / 1071
Master                   324 / 477
Enlightened          1217 / 1796
Burned                       - / 3085

### LINKED ISSUE #403: Include “potential reviews” in graph
I find the graph of upcoming reviews very helpful, but slightly lacking in one way. Say I have a stack of 50 reviews coming up at 10am. Maybe 40 of those are old items where the next review (if I get them right) is well in the future, and 10 of them are new items where they’ll show up in my review stack one or more times again just today. But currently, before I’ve tackled that stack of 50, there’s no indication of those extra reviews I’ll have in a few hours.

I understand that there is a difference between reviews that are definitely coming and predicted reviews that will come only if you do prior reviews at the earliest possible time and get them right, but I think showing the latter (in a different colour bar, I’d think something fainter to suggest the uncertainty) would be nice. Basically, “here are the reviews you’ll be doing if you do the definitely-scheduled ones at the earliest possible time and get them correct”. So if you wait and let reviews pile up, these prediction bars will also just shift back in time.

UInt2048:
Implements the first half of #112 by adding the list, and fixes #274 by breaking down the review count in the list by SRS (which can be toggled to item type)

The list is opened by tapping the graph, as if for more details.

### DIFF
--- a/ios/UpcomingReviewsChartItem.swift
+++ b/ios/UpcomingReviewsChartItem.swift
@@ -40,12 +40,12 @@ class UpcomingReviewsXAxisValueFormatter: IAxisValueFormatter {
 class UpcomingReviewsChartItem: NSObject, TKMModelItem {
   let upcomingReviews: [Int]
   let currentReviewCount: Int
-  let target: MainViewController
+  let target: NSObject
   let action: Selector
   let date: Date
 
   @objc init(_ upcomingReviews: [Int], currentReviewCount: Int, at date: Date,
-             target: MainViewController, action: Selector) {
+             target: NSObject, action: Selector) {
     self.upcomingReviews = upcomingReviews
     self.currentReviewCount = currentReviewCount
     self.date = date
@@ -64,7 +64,7 @@ class UpcomingReviewsChartItem: NSObject, TKMModelItem {
 
 class UpcomingReviewsChartCell: TKMModelCell {
   private let view: CombinedChartView
-  private var mainVC: MainViewController!
+  private var targetController: NSObject!
   private var action: Selector!
 
   override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
@@ -102,7 +102,7 @@ class UpcomingReviewsChartCell: TKMModelCell {
 
   override func update(with baseItem: TKMModelItem!) {
     let item = baseItem as! UpcomingReviewsChartItem
-    mainVC = item.target
+    targetController = item.target
     action = item.action
 
     var hourlyData = [BarChartDataEntry]()
@@ -149,6 +149,6 @@ class UpcomingReviewsChartCell: TKMModelCell {
   }
 
   override func didSelect() {
-    mainVC.perform(action)
+    targetController.perform(action)
   }
 }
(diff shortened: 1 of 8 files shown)


UInt2048:
@davidsansome Ready for review.

ios/UpcomingReviewsChartItem.swift
```diff
@@ -39,12 +40,17 @@ class UpcomingReviewsXAxisValueFormatter: IAxisValueFormatter {
 class UpcomingReviewsChartItem: NSObject, TKMModelItem {
   let upcomingReviews: [Int]
   let currentReviewCount: Int
+  let target: MainViewController
```
davidsansome, line 43: Change the type of this to `NSObject` or `NSObjectProtocol` to avoid the tight coupling

ios/UpcomingReviewsViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+  // MARK: - UIView
+
+  override func viewWillAppear(_ animated: Bool) {
+    super.viewWillAppear(animated)
+    navigationController?.isNavigationBarHidden = false
+    date = UpcomingReviewsDateFormatter(Date())
+    rerender()
+  }
+
+  private typealias ReviewData = ReviewComposition
```
davidsansome, line 47: What's this type alias for?

ios/UpcomingReviewsViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+
+  private func combineData(_ a: ReviewData, _ b: ReviewData) -> ReviewData {
```
davidsansome, line 49: This might make more sense as a method (or a + operator overload) on the ReviewComposition struct itself

davidsansome (approved):

ios/UpcomingReviewsChartItem.swift
UInt2048: Done.

ios/UpcomingReviewsViewController.swift
UInt2048: No longer needed, thanks!

ios/UpcomingReviewsViewController.swift
UInt2048: You're right. I forgot Swift had operator overloading... that's been fixed!

UInt2048:
@davidsansome I've implemented the requested changes.

PR #478, opened 2021-07-09
Fix useless exact match & unshowable subject details

### LINKED ISSUE #86: Close but no cigar & disable partial answer match

### LINKED ISSUE #172: Emulating Wanikani behavior on wrong answer

### LINKED ISSUE #463: Meaning/Reading are randomly not shown
2054A130-1C8C-4EB1-BC77-A2CB9090E212

Since the last update, occasionally, the reading explanation or the meaning explanation is not being shown. (As it can be seen in the image, only the reading explanation is being displayed)

SafePointerT:
Due to bugs, the exact match setting does nothing, the script to change bundle identifers incorrectly matches the doctypes in the complication property lists, and if you review and get the reading wrong but haven't answered the meaning, or vice versa, there's no way to show the meaning and its explanation.

By fixing these bugs, this PR will resolve #86, resolve #172, and resolve #463.

### DIFF
--- a/ios/SubjectDetailsView.swift
+++ b/ios/SubjectDetailsView.swift
@@ -135,6 +135,11 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
   private var tableModel: TKMTableModel?
   private var lastSubjectChipTapped: SubjectChip?
 
+  private var subject: TKMSubject!
+  private var studyMaterials: TKMStudyMaterials?
+  private var assignment: TKMAssignment?
+  private var task: ReviewItem?
+
   public func setup(services: TKMServices, delegate: SubjectDelegate) {
     self.services = services
     subjectDelegate = delegate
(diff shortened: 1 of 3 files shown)


ios/SubjectDetailsView.swift
```diff
@@ -326,22 +361,30 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
...
       addComponents(subject, title: "Kanji", toModel: model)
 
-      if !isReview || meaningAttempted {
+      if meaningShown {
         model.addSection("Meaning Explanation")
         addFormattedText(subject.vocabulary.meaningExplanation,
                          isHint: false, toModel: model)
       }
-      if !isReview || readingAttempted {
+      if meaningShown, readingShown {
```
davidsansome, line 381: How come the reading explanation is only shown if both meaningShown and readingShown are true?

If it's intentional can you add a comment to explain why?

davidsansome (approved):
Looks great - thanks for fixing these!

In future please try to separate each fix into its own pull request. It can get quite confusing to review otherwise!

ios/SubjectDetailsView.swift
SafePointerT: @davidsansome Yes. The reading explanation usually also spoils the meaning. For example, https://www.wanikani.com/kanji/%E4%B8%80

davidsansome: Ah ok makes sense! Could you put that in a `// Comment` above this line to make it clear to the next person to read the code?

SafePointerT: @davidsansome Absolutely, just added that!

UInt2048:
> In future please try to separate each fix into its own pull request

I'd like to apologize for this. I realize that I haven't exactly been doing this, and I'd like to fix that. I'm going to mark my unwieldy PRs as draft so that I can split them up. Could you please give me another chance, or tell me if I've assumed incorrectly?

davidsansome:
Yeah that PR is pretty much impossible to review right now :D If you can split each individual change/feature into separate PRs I'll definitely take another look!

UInt2048:
@davidsansome Could you please take a look at #470? It only affects a single line of code.

PR #479, opened 2021-07-11
Edit notes (Fixes #27, #361)

### LINKED ISSUE #27: Allow users to make notes on meanings

### LINKED ISSUE #361: Feature request: improve the user note feature

UInt2048:
Fixes #27 and closes #361 by adding edit meaning/reading notes buttons.

### DIFF
--- a/ios/SubjectDetailsView.swift
+++ b/ios/SubjectDetailsView.swift
@@ -160,6 +160,12 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
       model.addSection("Meaning Note")
       model.add(notesItem)
     }
+    let hasMeaningNote = studyMaterials?.hasMeaningNote != nil
+    model.add(TKMBasicModelItem(style: .default,
+                                title: "\(hasMeaningNote ? "Edit" : "Create") meaning note",
+                                subtitle: nil,
+                                accessoryType: .none, target: self,
+                                action: #selector(editMeaningNote)))
   }
 
   private func addReadings(_ subject: TKMSubject,
@@ -183,6 +189,12 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
       model.addSection("Reading Note")
       model.add(notesItem)
     }
+    let hasReadingNote = studyMaterials?.hasReadingNote != nil
+    model.add(TKMBasicModelItem(style: .default,
+                                title: "\(hasReadingNote ? "Edit" : "Create") reading note",
+                                subtitle: nil,
+                                accessoryType: .none, target: self,
+                                action: #selector(editReadingNote)))
   }
 
   private func addComponents(_ subject: TKMSubject,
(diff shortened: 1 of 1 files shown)


ios/SubjectDetailsView.swift
```diff
@@ -160,6 +160,12 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
       model.addSection("Meaning Note")
       model.add(notesItem)
     }
+    let hasMeaningNote = studyMaterials?.hasMeaningNote != nil
```
davidsansome, line 163: hasMeaningNote is a bool - I think testing for nil is gonna give you the wrong result when the bool is false.

I would also check if the string is empty to catch the case when the proto field is set to the empty string.

`let hasMeaningNote = !(studyMaterials.meaningNote ?? "").isEmpty`

ios/SubjectDetailsView.swift
```diff
@@ -183,6 +189,12 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
       model.addSection("Reading Note")
       model.add(notesItem)
     }
+    let hasReadingNote = studyMaterials?.hasReadingNote != nil
```
davidsansome, line 192: Same here

ios/SubjectDetailsView.swift
```diff
@@ -431,6 +443,44 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
...
+    editNote(subject: subject, studyMaterials: studyMaterials, isMeaningNote: false)
+  }
+
+  func editNote(subject: TKMSubject, studyMaterials: TKMStudyMaterials?,
+                isMeaningNote: Bool) {
+    var materials = studyMaterials ?? TKMStudyMaterials()
+    if studyMaterials == nil {
+      materials.subjectID = subject.id
+    }
+    let hasNote = isMeaningNote ? materials.hasMeaningNote : materials.hasReadingNote
```
davidsansome, line 460: Make sure the logic is the same here too

davidsansome (approved):

ios/SubjectDetailsView.swift
UInt2048: Thanks for the tip! I've updated these three checks as requested.

PR #480, opened 2021-07-11
Add toggle for newly fixed exact match setting so users can set it

UInt2048:
The bug fixes in #478 seem to have accidentally forgotten to add the exact match setting to SettingsViewController. This PR corrects that.

### DIFF
--- a/ios/SettingsViewController.swift
+++ b/ios/SettingsViewController.swift
@@ -147,6 +147,12 @@ class SettingsViewController: UITableViewController {
                                 action: #selector(fontSizeChanged(_:))))
 
     model.add(TKMSwitchModelItem(style: .subtitle,
+                                     title: "Exact match",
+                                     subtitle: "Requires typing in answers exactly correct",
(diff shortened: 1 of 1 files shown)


ios/SettingsViewController.swift
```diff
@@ -146,6 +146,12 @@ class SettingsViewController: UITableViewController {
                                 target: self,
                                 action: #selector(fontSizeChanged(_:))))
 
+    model.add(TKMSwitchModelItem(style: .subtitle,
+                                     title: "Exact match",
```
davidsansome, line 150: Can you fix the indentation here? I wonder why swift-format isn't doing it.

UInt2048: Sorry about that! When I split up the PRs, I copied and pasted the code from the original in some cases. I guess I didn't get SwiftFormat to fix that part.

davidsansome: Ah ok no problem!

PR #483, opened 2021-07-11
Prepare for list view by providing a detailed review composition

UInt2048:


ios/LocalCachingClient.swift
```diff
@@ -26,6 +26,30 @@ extension Notification.Name {
...
+      kanji: Int,
+      vocab: Int,
+      apprentice: Int,
+      guru: Int,
+      master: Int,
+      enlightened: Int
+}
+
+extension ReviewComposition {
+  init() {
```
davidsansome, line 41: Rather than initialising all these separately, can't do you do it in the struct definition?

```
struct ReviewComposition {
  var reviews = 0
  var radical = 0
  ...
}

File: ios/LocalCachingClient.swift [hunk 2]
```diff
@@ -26,6 +26,30 @@ extension Notification.Name {
   static let lccSRSCategoryCountsChanged = Notification.Name("lccSRSCategoryCountsChanged")
 }
 
+struct ReviewComposition {
+  var reviews: Int,
```
davidsansome, line 30: I would name this one `count` or `totalCount` or `totalReviews`

ios/LocalCachingClient.swift
```diff
@@ -26,6 +26,30 @@ extension Notification.Name {
...
+      radical: Int,
```
davidsansome, line 31: I wonder if it would be cleaner to use two maps here instead:

```
var countByType: [TKMSubject.TypeEnum: Int] = [:]
var countByCategory: [SRSStageCategory: Int] = [:]
```

ios/LocalCachingClient.swift
```diff
@@ -122,15 +145,28 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
...
-    var reviewCount = 0
-    var upcomingReviews = Array(repeating: 0, count: 48)
+    var lessonCount = 0,
+      reviewComposition = Array(repeating: ReviewComposition(), count: SRSStage.maxHours + 1)
+
+    func iterateValidReview(_ type: TKMSubject.TypeEnum, hours: Int, stage: SRSStage) {
+      guard hours < reviewComposition.count else { return }
+      reviewComposition[hours].reviews += 1
+
+      if type == .radical { reviewComposition[hours].radical += 1 }
```
davidsansome, line 161: If you use maps this becomes simpler:

```
reviewComposition.countByType[type, default: 0] += 1
reviewComposition.countByCategory[stage.category, default: 0] += 1
```

ios/WaniKaniAPI/Sources/WaniKaniAPI/ProtobufExtensions.swift
```diff
@@ -130,6 +130,8 @@ public enum SRSStage: Int, CustomStringConvertible, Comparable, Strideable {
     case .burned: return .burned
     }
   }
+  
+  public static let maxHours = 2879
```
davidsansome, line 134: Can you add a comment to say what 2879 is (and why)?

davidsansome (approved):

ios/LocalCachingClient.swift
UInt2048: That's the more efficient way I was looking for, thanks!

ios/LocalCachingClient.swift
UInt2048: I changed the name to `availableReviews` - it's the number of reviews that become (or, in the case of the current hour, are) available.

ios/LocalCachingClient.swift
UInt2048: Good idea. I've implemented this.

ios/LocalCachingClient.swift
UInt2048: Yep, definitely cleaner!

ios/WaniKaniAPI/Sources/WaniKaniAPI/ProtobufExtensions.swift
UInt2048: Sure, I've added this.

PR #486, opened 2021-07-13
Add toggle to disable note editing in reviews

UInt2048:


davidsansome:
Why do we need a toggle for it? It seems like a useful feature for everyone!

If it's because the button is too big and distracting, maybe we can rework it into a smaller "edit" icon button on the side of the UI?

UInt2048:
@davidsansome Yes, exactly, it's too big, so it can interfere with Anki mode if you're not using back-to-back reading before meaning.

I like the idea of a smaller "edit" button on the side, but I'm not sure how to implement that at the moment.

davidsansome:
Ah alright, I'll merge this now and I might have a go at doing the small edit button later.

I think we need to rework this list of settings at some point too - it's getting pretty long!

PR #493, opened 2021-07-14
Meaning only back-to-back order for lightning Anki reviews

UInt2048:
Intended to make Anki mode reviews faster, this draft pull request proposes to add a back-to-back option to only test the meaning, so people can answer whether they got both the meaning and reading correct in one go to save button presses.


UInt2048:
Closing draft PR, will reopen if ready.

PR #502, opened 2021-07-17
Regenerate Protobuf file that shouldn't be formatted

UInt2048:
Somehow, upon moving files to WaniKaniAPI, `wanikani_api.pb.swift` accidentally got formatted by SwiftFormat.

This PR proposes to fix that by regenerating it from the original Protobuf file.


davidsansome:
Interesting, I would've thought the `--exclude **.pb.swift` in https://github.com/davidsansome/tsurukame/blob/master/ios/.swiftformat would've prevented that.

Oh well, thanks for fixing it!

UInt2048:
@davidsansome Somehow it didn't when you were moving things into the WaniKaniAPI module. It won't get formatted on building Tsurukame.

PR #503, opened 2021-07-19
Add burned field to stats section

UInt2048:
This PR allows users to see the date they have burned an item in the stats section.

For future reference, in order for the burned date to be cached into the client the first time, users would have to perform a full refresh (since caches created prior, of course, wouldn't contain the burned at date).


UInt2048:
This PR, and the ones created afterwards, are ready to be reviewed.

PR #519, opened 2021-10-15
Switch some IDs to Int64

### LINKED ISSUE #517: Missing reviews
Hi there,
My reviews aren’t updating unless I uninstall and reinstall the app. It just started happening, other than that the pp works great!

### LINKED ISSUE #518: Tsurukame Will Not Fully Sync to Web
As the title implies, I'm having issues getting Tsurukame to sync up fully with the web version of WaniKani.

I did some reviews and synced normally before boarding a plane to Japan; on landing and connecting to the hotel wifi, I'm finding that my ~170 reviews remaining in-app are not reflected online, which claims I have 260 to do. Interestingly, _something_ appears to be syncing: pulling down to refresh and then checking online will sometimes drop my online reviews down a few, to 259 or something, but never all the way to the "true" number.

This occurs on both the stable and Testflight versions of the app.

tellowkrinkle:
Review IDs are now over 2 billion!!! (AKA Int32.max).  I upgraded the IDs of things that look dynamically generated (and therefore possible to overflow, even if it won't happen for a long time) to Int64 in the protobuf.  AFAIK this is a backwards compatible change (one way, the newer protobuf can read the old one but not the other way around).

For the other ones, I did a checked, crashing conversion to Int32 (so if WK does decide to issue kanji or level 2^31, it'll crash the app).  Would you prefer swapping to nil or something else instead?

Fixes #518
Fixes #517


ios/WaniKaniAPI/Sources/WaniKaniAPI/APIClient.swift
```diff
@@ -158,7 +158,7 @@ public class WaniKaniAPIClient: NSObject {
     }.map { (allData: Response<[Response<LevelProgressionData>]>) -> LevelProgressions in
       var ret = [TKMLevel]()
       for data in allData.data {
-        ret.append(data.data.toProto(id: data.id))
+        ret.append(data.data.toProto(id: data.id.map(Int32.init)))
```
davidsansome, line 161: Let's convert all the IDs to Int64s for consistency. So here I think you'd need to change the type in LevelProgressionData and TKMLevel.

tellowkrinkle:
All switched over

PR #524, opened 2021-11-07
Enable App Group capability

UInt2048:
This PR simply enables the app group capability, a requirement for future PRs such as for a widget.

If App Groups aren't enabled for your provisioning profile in App Store Connect, you'll need to do so.

PR #530, opened 2021-11-24
Rename tests bundle identifier

LittenArchive:
This PR renames the tests bundle identifier to be consistent with the `com.davidsansome.wanikani` used for other bundle identifiers. It seems to be a relic of the original name of the app.

P.S. When you review this, could you also review #524? It's also really small.


UInt2048:
@davidsansome Could you please look into this PR? It's only a diff of 2 lines.

PR #576, opened 2022-07-28
fix section header clipping the right way

indirect:
thanks to a friend at apple, it was pointed out to me that the storyboard xib had hardcoded section heights, apparently originally due to a long-ago bug in xcode. removing the incorrect hardcoded heights from the xib means my weird hack-around isn't needed anymore! phew.


davidsansome:
Oh nice, thanks for continuing to look into this!

PR #588, opened 2022-09-22
remove keyboard inset hack, unglitch iPhone 14 Pro

### LINKED ISSUE #587: Bug: Top of meaning/reading page is cut off on new iPhone models
As can be seen in these pictures. I’m on an iPhone 14 Pro. This happens after getting something wrong.

2C03A2E7-858F-4EAE-81DF-B665F4360881

E0E994E1-6242-429C-BBE5-4C8D1D7C1F0A

indirect:
As seen in screenshots on #587, the UITextField for the answer sometimes seems to move itself down without adjusting the color gradient above it or the lesson view below it. in my testing, this seems to happen when switching between english and japanese keyboards automatically, which sometimes (but not always) triggers the inset hack. After the inset hack has triggered, the heights seem to be wrong.

I'm not sure why this seems to trigger so much more often on iPhone 14 Pro, and not before, but deleting the inset hack seems to resolve this issue, in my testing on a physical iPhone 14 Pro.

Video captured from build equivalent to ad3b00857aba42cd3cb9e4cb4945410bd3d01a52, with reproducible glitch:

https://user-images.githubusercontent.com/78/191701479-6654fce2-d2a8-435e-8ed4-fa4e23c57ce4.mp4

Video captured from build equivalent to f00d0c467bab72a6424faf1d357909cdd680ae18, no more glitch:

https://user-images.githubusercontent.com/78/191701566-23438a7c-e43b-4651-a46c-32ea9642b0bd.mp4

fixes #587 (I hope)

### DIFF
--- a/ios/ReviewViewController.swift
+++ b/ios/ReviewViewController.swift
@@ -183,7 +183,6 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, SubjectDelega
   // These are set to match the keyboard animation.
   private var animationDuration: Double = kDefaultAnimationDuration
   private var animationCurve: UIView.AnimationCurve = kDefaultAnimationCurve
-  private var previousKeyboardInsetHeight: CGFloat?
 
   private var currentFontName: String!
   private var normalFontName: String!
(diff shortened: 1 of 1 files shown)


davidsansome:
Thanks for the PR! I don't want to remove this hack completely - it's supposed to keep the question label in the same place when the keyboard makes small size changes, like switching language or showing the suggestions bar.

Instead I've pushed 2cf2de5cccc82052eee681d4d0dd4e0a158c3264 which correctly pins the background gradient to the bottom. I'm not sure why this worked prior to iOS 16!

PR #598, opened 2022-11-07
Account for resets when estimating level time

indirect:
When a Wanikani user resets to a lower level, the level they were on when they reset never gets a `completedAt`, it gets `abandonedAt` instead. That means we have been including the entire time from the moment each abandoned/reset level was started until the present moment in the average calculation—for me personally, that meant the estimated level times were around 90 days. Excluding level time records that have an abandonedAt present fixes the estimation. While reviewing the calculation, I noticed that we were sometimes including zeroes in the list of level times, so I added a check to remove those.

Before: Estimated level time 90d
After: Estimated level time 10d

Since I have never taken 90 days to complete a level, and usually take about 10 days, it seems like that fixed it.


davidsansome:
Oooh I'd noticed this too but hadn't investigated what was causing it. Thank you!

PR #605, opened 2023-01-02
Allow API token login

### LINKED ISSUE #571: Feature Request: Allow Sign In with API Token only, no need to provide Username and Password
Has adding support for signing in with an API token been discussed anywhere? I wasn't able to find it being previously mentioned.

Thank you 😅

Deadpikle:
Adds alternative login method using API token rather than username/password credentials. User has the option to use one login method or the other, so this should not turn people away from using the app at all.

I did have to tweak the layout of the main login screen to accommodate the extra height. Also tweaked the shadow to make it more readable since the main Sign In text moved into a lighter background, but that's arguably opinionated and can be easily removed. See below screenshot.

Downside: User email address cannot be obtained via the API, thus making the gravatar fetch not work. In theory, we could add a setting that would allow users to override their email address used for Gravatar pulling. Is this a good option, or is there another way to grab the user's email address without web scraping that I missed...?

simulator_screenshot_CA922CBE-1CAF-4E92-B9D4-63FE9DF04CEB

Closes #571

### DIFF
--- a/ios/AppDelegate.swift
+++ b/ios/AppDelegate.swift
@@ -45,7 +45,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, LoginViewControllerDelega
     let nc = NotificationCenter.default
     nc.addObserver(self, selector: #selector(logout), name: .logout, object: nil)
 
-    if !Settings.userApiToken.isEmpty, !Settings.userCookie.isEmpty {
+    if !Settings.userApiToken.isEmpty {
       setMainViewControllerAnimated(animated: false, clearUserData: false)
     } else {
       pushLoginViewController()
@@ -132,7 +132,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, LoginViewControllerDelega
 
   @objc
   private func logout(_: Notification?) {
-    Settings.userCookie = ""
     Settings.userApiToken = ""
     Settings.userEmailAddress = ""
     services.localCachingClient.clearAllDataAndClose()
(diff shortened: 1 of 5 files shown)


davidsansome (changes requested):
Thanks for the PR!

Some general comments:

1. The login page looks pretty busy now. How about simplifying it by adding a button that toggles between the two login mechanisms. It could be a text-link-style button, like the Privacy Policy one at the very bottom.

    Initially it would look like:
    - "Username" field
    - "Password" field
    - "Sign in" button
    - (some vertical space)
    - "Use API token" text-link button

    Pressing the toggle would change it to:
    - "API token" field
    - "Sign in" button
    - (some vertical space)
    - "Use API username and password" text-link button

1. We should call them "API token"s rather than "API key"s everywhere, to match the terminology on the WaniKani site.
1. Can you test what happens if the user uses a key that doesn't have all the required permissions? I don't think we can detect that at login time, so we'd need a detailed, actionable error dialog when we get an error response from the API. Ideally it would have a link that opens `https://www.wanikani.com/settings/personal_access_tokens#modal_{{ token }}` in Safari.

Deadpikle:
Your comments have been addressed. Thank you for the review. 😄 Here's what it looks like currently:

api-login-swap

> Can you test what happens if the user uses a key that doesn't have all the required permissions?

Yeah...this is a little weird. The API just returns a `401` just the same as we'd expect from an expired token with a string message that the API token doesn't have the right permissions. I didn't really want to rely on the text of the returned error message to know if it was an expired token, token without permissions, etc. since text can change at any time. Given that I wasn't aware of a better option than parsing the text, I went ahead and just updated the current token expiration modal with more information and a direct link to the WaniKani website for tokens. I figure that a user who is logging in with an existing API token will know what to do with the given information, so I believe that will suffice. Here's what that looks like:

Simulator Screen Shot - iPhone SE (2nd generation) - 2023-01-22 at 10 45 55

davidsansome:
My turn for the slow reply this time 😄 

Thanks again for doing this. I've added a commit to tidy up a few things and add a paste button next to the API key field.

Issue #219, opened 2019-11-29
Feature Request: Half-Level Setting


turtlemaster19:
A user can level up, but has vocab reviews stopping them from actually starting the next level. Can there be a setting where instead of showing the number of the new level, it shows a half-level?

For example, I officially levelled up to Level 19 a few days ago. During the period that I was doing some vocab lessons, I was actually "Level 18.5," but today I did my radical lessons so I'm now "Level 19."

zebala:
Similarly I would love to be able to see my progress on the previous level, since that is what I'm actually interested in. I don't do items from the new level until I have finished the vocab, so it is just pointless to show the progress circles of the new level.

mzsanford:
I've also toyed with other representations<sup>1</sup> in my design work for #119.  If this is something that the app is going to support then these notes might be helpful:

Given the specific space constraints of Apple Watch complications in #119 something like "18↑" might work better than "18.5". It feels like "18↗︎" might be an even better representation but I have yet to test the relative width and font support. Other options I've considered are leading arrows like "→19" and "↗︎19" but I've also not tested how these _feel_ in the UI.

Overall this seems like an issue with the notion of levels on the WaniKani side that is better fixed on the whole site.

<hr/>
<sup>1</sup> I started with "18*" or "19*" but it didn't seem clear – I really hate a * without an explanation.

turtlemaster19:
@mzsanford There's always ½

turtlemaster19:
@mzsanford What do you think?

PR #606, opened 2023-01-02
Add option for viewing current level graph before finishing vocab

Deadpikle:
One downside to passing kanji is that you level up before you finish vocabulary. For those of us who are ordering things such that we finish vocab before we start the next level's radicals/kanji, it would be useful to see a graph of our current progress.

This PR adds a setting to enable that. If active, it will check to see if the user's previous level (the one where they have vocabulary) is done or not. If it is not done, a second graph is shown on the home screen with the title `Current Level (#)`, and then the user's _actual_ current level graph is titled with `Next Level`. If the setting is on and the user has finished the prior level, it will only show the current level's graph.

I also renamed `This Level` to `Current Level` if the setting is off (or does not apply); that is more of a subjective change that doesn't necessarily need to come in. I just thought it would make sense if we always have "current level" and "next level", as if the user is done with the prior level, the section title would change from "Next Level" to "This Level", which might end up being confusing.

Simulator Screen Shot - iPhone SE (2nd generation) - 2023-01-02 at 12 00 12

Simulator Screen Shot - iPhone SE (2nd generation) - 2023-01-02 at 12 00 19

Related to: #219

### DIFF
--- a/ios/LocalCachingClient.swift
+++ b/ios/LocalCachingClient.swift
@@ -502,16 +502,6 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
     return !assignments.contains { !$0.hasPassedAt }
   }
 
-  func getAssignmentsAtUsersCurrentGraphLevel() -> [TKMAssignment] {
-    guard let userInfo = getUserInfo() else {
-      return []
-    }
-    if Settings.showPreviousLevelGraph, !hasCompletedPreviousLevel() {
-      return getAssignments(level: Int(userInfo.level) - 1)
-    }
-    return getAssignments(level: Int(userInfo.level))
-  }
-
   private func getAssignments(level: Int, transaction db: FMDatabase) -> [TKMAssignment] {
     var ret = [TKMAssignment]()
     var subjectIds = Set<Int>()
--- a/ios/LocalCachingClient.swift
+++ b/ios/LocalCachingClient.swift
@@ -498,13 +498,8 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
       level = level - 1
     }
     let assignments = getAssignments(level: level)
-    var isDoneWithPreviousLevel = true
-    for assignment in assignments {
-      if assignment.srsStageNumber == level, !assignment.hasPassedAt {
-        isDoneWithPreviousLevel = false
-      }
-    }
-    return isDoneWithPreviousLevel
+    // if any assignment has not been passed, they have not passed the prior level
+    return !assignments.contains { !$0.hasPassedAt }
   }
 
   func getAssignmentsAtUsersCurrentGraphLevel() -> [TKMAssignment] {
--- a/ios/LocalCachingClient.swift
+++ b/ios/LocalCachingClient.swift
@@ -489,6 +489,34 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
     return getAssignments(level: Int(userInfo.level))
   }
 
+  func hasCompletedPreviousLevel() -> Bool {
+    guard let userInfo = getUserInfo() else {
+      return false
+    }
+    var level = Int(userInfo.level)
+    if level > 1 {
+      level = level - 1
+    }
+    let assignments = getAssignments(level: level)
+    var isDoneWithPreviousLevel = true
+    for assignment in assignments {
+      if assignment.srsStageNumber == level, !assignment.hasPassedAt {
+        isDoneWithPreviousLevel = false
+      }
+    }
+    return isDoneWithPreviousLevel
+  }
+
+  func getAssignmentsAtUsersCurrentGraphLevel() -> [TKMAssignment] {
+    guard let userInfo = getUserInfo() else {
+      return []
+    }
+    if Settings.showPreviousLevelGraph, !hasCompletedPreviousLevel() {
+      return getAssignments(level: Int(userInfo.level) - 1)
+    }
+    return getAssignments(level: Int(userInfo.level))
+  }
+
   private func getAssignments(level: Int, transaction db: FMDatabase) -> [TKMAssignment] {
     var ret = [TKMAssignment]()
     var subjectIds = Set<Int>()
--- a/ios/MainViewController.swift
+++ b/ios/MainViewController.swift
@@ -201,7 +201,7 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
       model.add(section: "Current level (\(user.currentLevel > 1 ? user.currentLevel - 1 : 1))")
       let currentGraphLevelAssignments = services.localCachingClient.getAssignments(level: Int(user.currentLevel) - 1)
       model.add(CurrentLevelChartItem(currentLevelAssignments: currentGraphLevelAssignments))
-      model.add(section: "Next level")
+      model.add(section: "Next level (\(user.currentLevel))")
     } else {
       model.add(section: "Current level")
     }
--- a/ios/MainViewController.swift
+++ b/ios/MainViewController.swift
@@ -197,10 +197,9 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
                                               currentLevelAssignments: currentLevelAssignments))
     }
 
-    if Settings.showPreviousLevelGraph, !services.localCachingClient.hasCompletedPreviousLevel() {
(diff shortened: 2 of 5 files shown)


ios/LocalCachingClient.swift
```diff
@@ -489,6 +489,34 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
...
+    }
+    var level = Int(userInfo.level)
+    if level > 1 {
+      level = level - 1
+    }
+    let assignments = getAssignments(level: level)
+    var isDoneWithPreviousLevel = true
+    for assignment in assignments {
+      if assignment.srsStageNumber == level, !assignment.hasPassedAt {
+        isDoneWithPreviousLevel = false
```
davidsansome, line 504: You can exit early here, just `return false` and ditch the `isDoneWithPreviousLevel` variable.

Alternatively, if you think it's more readable, you could try something like:

```
return !assignments.contains { !$0.hasPassedAt }
```

ios/LocalCachingClient.swift
```diff
@@ -489,6 +489,34 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
...
+      return false
+    }
+    var level = Int(userInfo.level)
+    if level > 1 {
+      level = level - 1
+    }
+    let assignments = getAssignments(level: level)
+    var isDoneWithPreviousLevel = true
+    for assignment in assignments {
+      if assignment.srsStageNumber == level, !assignment.hasPassedAt {
```
davidsansome, line 503: The `srsStageNumber` is the Apprentice, Guru... stage number, **not** the user's level. You can't compare the two like this.

If you're trying to check whether the assignment is past Guru, you can do `if assignment.srsStage >= .guru1`.

But, is that even necessary since we're checking `hasPassedAt` anyway?

ios/LocalCachingClient.swift
```diff
@@ -489,6 +489,34 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
...
+        isDoneWithPreviousLevel = false
+      }
+    }
+    return isDoneWithPreviousLevel
+  }
+
+  func getAssignmentsAtUsersCurrentGraphLevel() -> [TKMAssignment] {
```
davidsansome, line 510: I'm not sure this function is necessary. MainViewController is doing all these checks anyway, so let's just delete this function and make it do `services.localCachingClient.getAssignments(level: user.currentLevel - 1)`

ios/MainViewController.swift
```diff
@@ -197,7 +197,16 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
                                               currentLevelAssignments: currentLevelAssignments))
     }
 
-    model.add(section: "This level")
+    if Settings.showPreviousLevelGraph, !services.localCachingClient.hasCompletedPreviousLevel() {
+      model.add(section: "Current level (\(user.currentLevel > 1 ? user.currentLevel - 1 : 1))")
+      let currentGraphLevelAssignments = services.localCachingClient
+        .getAssignmentsAtUsersCurrentGraphLevel()
+      model.add(CurrentLevelChartItem(currentLevelAssignments: currentGraphLevelAssignments))
+      model.add(section: "Next level")
```
davidsansome, line 205: It might be nice to put the next level number in parentheses here too.

ios/MainViewController.swift
```diff
@@ -197,7 +197,16 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
...
+    } else {
+      model.add(section: "Current level")
+    }
+
```
davidsansome, line 209: Can we have "Show remaining" and "Show all" buttons for the current level? It might need a bit of refactoring.

ios/Settings.swift
```diff
@@ -194,6 +194,7 @@ protocol SettingProtocol {
   @Setting(false, #keyPath(allowSkippingReviews)) static var allowSkippingReviews: Bool
   @Setting(true, #keyPath(minimizeReviewPenalty)) static var minimizeReviewPenalty: Bool
   @Setting(false, #keyPath(ankiMode)) static var ankiMode: Bool
+  @Setting(false, #keyPath(showPreviousLevelGraph)) static var showPreviousLevelGraph: Bool
```
davidsansome, line 197: I'd probably have this default to true tbh - it seems really useful, and people are unlikely to discover it otherwise.

davidsansome (changes requested):

ios/MainViewController.swift
Deadpikle: `SubjectCatalogueViewController` already took a level as a param so this didn't take much work.

`SubjectsRemainingViewController` was refactored to take the user's level and then pulling in vocabulary into the list if the setting for this PR (`showPreviousLevelGraph`) is `true`.

I thought about using some new segues in the storyboard to help know which level a user was tapping on, but decided that two new `var` in the `MainViewController` was much simpler and less confusing.

ios/LocalCachingClient.swift
Deadpikle: I went with the latter syntax you recommended. I'm (historically) an Objective-C user, and haven't used Swift really much at all, so I'll lean on your experience for the syntax there being widely used. (The `$0` syntax was understood after a quick google search, so shouldn't be a big deal.)

Deadpikle:
Thanks for the very quick review! (I'm usually fairly busy, so sometimes these things take me some time. If you want to handle anything in this or the other PR without waiting for me, please do! Otherwise, will try to be relatively quick.)

Here's a GIF of this working:

two-levels-at-once

ios/SubjectsRemainingViewController.swift
```diff
@@ -33,22 +35,20 @@ class SubjectsRemainingViewController: UITableViewController, SubjectDelegate,
   override func viewDidLoad() {
     super.viewDidLoad()
 
-    guard let level = services.localCachingClient.getUserInfo()?.level else {
-      return
-    }
-    navigationItem.title = "Remaining in Level \(level)"
+    navigationItem.title = "Remaining in Level \(level!)"
```
davidsansome, line 38: You probably don't need the `!` here, since the variable is defined as an `Int!` already.

ios/MainViewController.swift
```diff
@@ -61,6 +61,9 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
   var hasReviews = false
   var updatingTableModel = false
 
+  var isShowingPriorUserLevel = false
```
davidsansome, line 64: These two variables feel a bit icky to me. What about having a single `selectedSubjectCatalogLevel`, and setting it in the tap handler.

`addShowRemainingAllItems` would need to take a `level` parameter, and it can use the new syntax for `BasicModelItem` tap handlers, like this:

```
    model.add(BasicModelItem(style: .default,
                             title: "Show remaining",
                             subtitle: nil,
                             accessoryType: .disclosureIndicator) {
      this.selectedSubjectCatalogLevel = level
      performSegue(withIdentifier: "showRemaining", sender: self)
    })
```

Then you can delete the `showRemaining` and `showAll` functions as well.

davidsansome:
> Thanks for the very quick review! (I'm usually fairly busy, so sometimes these things take me some time. If you want to handle anything in this or the other PR without waiting for me, please do! Otherwise, will try to be relatively quick.)

No worries, thanks for addressing the comments!

> Here's a GIF of this working:

Looks great :)

> I'm not sure why the "Next level-up review" is set to "Now". Is that expected behavior? Doesn't seem to be affected by anything in this PR, though, so I'll write that off for the moment.

Hmm I'm not sure, is it because you still have 1 kanji left in level 3, and it's ready for review now? I would've thought it should be looking at the level 4 list though.

ios/SubjectsRemainingViewController.swift
Deadpikle: I see exactly what you mean here, and I would imagine you'd be correct, but removing the `!` results in [this warning](https://stackoverflow.com/questions/42543007/how-to-solve-string-interpolation-produces-a-debug-description-for-an-optional). So, I opted for removing the `!` and just doing `private var level: Int = 0`. ....not sure if that is any better. I'll leave that one for your judgment call.

I'm on Xcode 12.2 with no real option to update right now, which is unfortunate, as that may be causing the problem (grumble grumble, why are Xcode updates so terribly large). Sorry about that.

Deadpikle:
Again, apologies for the slow here. Life and other priorities. Let me know if this or the other PR need changes or if they're good to go now. :)

PR #623, opened 2023-03-30
Fix iOS 12 crash

### LINKED ISSUE #619: Crash when opening the app
I have an iPhone SE with iOS version 12.2. I updated to the most recent release available on the App Store, and now I am unable to use the app. Whenever I open it, it crashes within a couple seconds. I was able to open it in airplane mode, but the app would crash as soon as I connected my phone to the internet. I tried reinstalling the app, and it is fine with the login screen, but as soon as I log in it crashes immediately.

I have the local-cache.db from before I reinstalled the app if that helps.

UInt2048:
Fix #619 with a one-line fix to a polyfill for iOS 12.

I have tested this PR in the iOS 12 simulator and confirmed it fixes the crash.


davidsansome:
Thanks!

UInt2048:
@davidsansome No problem!

You might need to bump the version to 1.27 for the TestFlight build to pass though ;)

davidsansome:
Ah yep I noticed that the other day but didn't get around to fixing it.
I've just done it now!

Issue #636, opened 2023-05-24
App freezes when doing lesson for する, a new kana-only


milowyner:
The app seems to be handling the new kana-only vocabulary fine except for する. When I tap it or swipe in its direction in the lessons the app freezes and no touch inputs do anything until I force close and relaunch. I’m on iOS 16.5 using an iPhone 12 Mini in case that helps.

matthewgrimes:
Same for me. Also on an iPhone, a 12 pro running 16.4.1. 

I can also add that the other lessons went through but now する is the first lesson and clicking on the lesson tab from the home screen freezes the app.

Twiglet1022:
Same here but with これ. As a temporary workaround I was able to get past it by installing the TestFlight beta (which also freezes for that word) and disabling kana-only vocab lessons.

iPhone 14 Pro iOS 16.4.1 (a)

MattMisrak:
Same here first with これ and then with する after closing the app a few times and a few full refresh. 

iPhone 12 iOS 16.4.1(a)

marciska:
Might be because する verbs are specially handled when highlighting example sentences.
I suppose (!) that there might be something wrong with these checks now.
Searching for "suruverb" shows up [quite a few locations](https://github.com/search?q=repo%3Adavidsansome%2Ftsurukame+suruverb&type=code).

Klhhff:
Same problem here. これ and する

marciska:
I pushed a hotfix that solves the issue (for now). Let's hope it gets quickly integrated into the base branch!

It is only a temporary fix: making all suruVerb of class ichidanVerb. This prevents the app from crashing, but example sentences for other sure-verbs are coloured wrong. Exchanging a critical crashing-bug with “only” a visual appearance bug, so to speak. At least until a real fix is proposed.

davidsansome:
I've updated the TestFlight build with a fix for する. I wasn't able to reproduce the problem with これ though.

Would somebody be able to install 1.28 (972) from TestFlight and verify it works for both now?

milowyner:
Yep, する is now working on the latest TestFlight, and これ was previously working for me as well. Thanks for the quick fix!

davidsansome:
Thanks for confirming! I'll push an update to the App Store now.

davidsansome:
I'll leave this bug open until the fix rolls out so people can find it.

The new release is waiting for App Store review now.

davidsansome:
[...]

PR #637, opened 2023-05-25
Critical Bugfix: crash when する appears

marciska:
Hotfix for #636.

Added an additional if statement so app doesn't crash when する appears.

Issue before was that する is recognised as a `suruVerb`, but is omitted since it is a suffix. In the code, there is then an empty string, resulting in a crash. The above fix checks if it is the kana-word する, and then does NOT omit it.


davidsansome:
Thanks for doing this! I've pushed 304f02a50a73e74d3ab8ec79f2f39194e6a255b2 instead which is I think a more general solution to the bug!

PR #640, opened 2023-05-30
Add ability to show minutes for upcoming level up review

bqmoreland:
The current default behavior of not showing minutes means that as a user I have to add the minutes on in my head. This setting removes that mental load if the user so desires.


davidsansome:
Thanks for the PR, and sorry for the super later response!

PR #644, opened 2023-06-18
Feature: Optional secondary review order sort by review subject type. 

theFestest:
This introduces an optional secondary review sort order by the review subject type (i.e. radical, kanji, vocabulary). The existing (unsorted) behavior is maintained by default.

Tested in simulator and via macOS Catalyst app.

I’m happy to make adjustments as you see fit. 😄

### DIFF
--- a/ios/ReviewSession.swift
+++ b/ios/ReviewSession.swift
@@ -204,6 +204,46 @@ class ReviewSession {
     _ = services.localCachingClient?.updateStudyMaterial(activeStudyMaterials!)
   }
 
+  private func sortReviewType() {
+    // Settings.secondaryReviewOrder -> (radical, kanji, vocab) order ranks
+    var subjectOrder = [UInt: (Int, Int, Int)]()
+    // Instead of more complex logic, let's have structed data and few sorting cases
+    subjectOrder[SecondaryReviewOrder.unsorted.rawValue] = (0, 0, 0) // Default, no sort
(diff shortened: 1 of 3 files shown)


tallerasaf:
@theFestest If I am choosing for example, review order: "Oldest available first"
And then the secondary review order: [radical, kanji, vocabulary]
It means the order of reviews will be 'radicals' sorted from oldest available first, then 'kanjis' sorted from oldest available first, then 'vocabularies' sorted from oldest available first?

So your new sort is actually the primary sort and the existing sort will be the secondary sort, right?

UInt2048:
@theFestest Feel free to use my code from #548 if you want.

Interesting to see your approach!

theFestest:
@tallerasaf  You make a good point. What you describe first does aligns with the behavior of this feature and I agree that the alternative would be more confusing. My goal was to provide the ability to choose the order of subject types to focus on in reviews like you say, for example doing all kanji first, and so on. 

The intention with saying “secondary” was really just that the new sorting is applied as a second pass and is optional.
 
Perhaps “Subject Order Sort” would be a more appropriate description for the feature.


@UInt2048 Thanks for the tip! 😄

tallerasaf:
@theFestest 

Basically you are copying the Lessons, "Order" feature into the review section:
image

So I think @davidsansome need to decide the convention, because right now the lessons "order" feature and the reviews "order" feature behaves differently.

I think consistency is really important, I suggest to rename the current lessons "order" to "Subject Order", and rename the current review "Order" to "Sort By", and then @theFestest new feature will be "Subject Order" the same as will be in the lessons section.

Issue #595, opened 2022-10-18
Info not syncing to Wanikani when in Anki mode


vanpiro:
i discovered that If I do reviews in anki mode (without typing the answers, after just a couple of times, many of the reviews i did will never sync with Wanikani. 
This does not happen when I am not in anki mode.

could you please fix this issue?

santialbo:
This happens when you go faster than the answer slide animations. If you go slower than the animations then the answers do count. I will upload a video tomorrow showing it.

santialbo:
Here's the video https://www.veed.io/view/43b04dac-0c60-465e-bd1e-c468ddad6d71

vanpiro:
Thanks
I figured out how to make sure it gets the credit of the review.

happy new year

santialbo:
Tagging @UInt2048, does this make sense? If you don't let the animations finish the answer won't be synced on Wanikani. It takes a full-refresh for tsurukame to realize. You can see a video showcasing the bug https://www.veed.io/view/43b04dac-0c60-465e-bd1e-c468ddad6d71

UInt2048:
Thanks for the video. I call this the negative one correct answers bug. Basically, Anki mode normally always marks your answer wrong (+1 incorrect) and then reverses it before submission (-1 incorrect). But if you manage to get that popup without the answer displaying, the answer is not marked wrong, and you get a free -1 correct.However, a submitted review to WaniKani telling it you got it incorrect -1 times is obviously rejected, and Tsurukame refreshes to find out that the review is still available.iPhoneから送信2023/01/04 3:30、Santi Albo ***@***.***>のメール:﻿
Tagging @UInt2048, does this make sense? If you don't let the animations finish the answer won't be synced on Wanikani. It takes a full-refresh for tsurukame to realize. You can see a video showcasing the bug https://www.veed.io/view/43b04dac-0c60-465e-bd1e-c468ddad6d71

santialbo:
If there is no easy way to fix this, could we at least make it so that you cannot tap the screen until the animation has finished?

UInt2048:
@santialbo Could you please build the commit 62e28a1 to see if that works?

santialbo:
I'm just a regular user of the app :S I don't know how to build it and test it.

marciska:
I tried testing it. Not sure if I did the correct thing, so take everything with a grain of salt. I checked out the commit 62e28a1, but it took me a while to get the simulator running on my Mac. I never worked with Xcode before.

[...]

PR #657, opened 2023-09-09
[Bugfix | Anki Mode] 595 - Sync Issues Causing Lost Progress

tarcisiotm:
Relevant Issue: https://github.com/davidsansome/tsurukame/issues/595

If the user taps the screen too fast after a correct submission, the just submitted **correct** answer will be marked **incorrect**, and this apparently gets rejected by Wanikani. I came a across this issue this week after I started using Anki mode, and got more than 200 reviews back when I forced a full sync, after suspecting something was off with Tsurukame's synchronization.

- fixed: Anki Mode tagging correct answers as incorrect when users tapped to show the next review element while the SubjectDetailsView was animating
- added: when users tap the canvas after clicking on a button (correct, incorrect, etc), but before the animation is done, the action gets cached, so the submit button will be automatically be processed once the animation ends

This is my first Pull Request here, so please let me know if something is not following the standards.

### DIFF
--- a/ios/ReviewViewController.swift
+++ b/ios/ReviewViewController.swift
@@ -174,6 +174,8 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, SubjectDelega
   private var session: ReviewSession!
 
   private var lastMarkAnswerWasFirstTime = false
+  private var ankiModeCachedSubmit = false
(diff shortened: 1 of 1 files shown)


davidsansome:
Cool, thanks for fixing this!

PR #659, opened 2023-09-13
Saw radical hotfix (Fixes #658)

### LINKED ISSUE #658: App crashes while reviewing certain radical
Hello,

I don’t know whether this is the correct place to post this but I’ve noticed app crashes every time the „saw“ radical comes up for review. 

I’m on iOS 16.6.1

UInt2048:
Don't crash if we don't actually have an image for the radical (in practice, it will show 📄 instead if not found).

If you'd prefer to crash, just don't commit the change to ProtobufExtensions.swift

Also, provide one for the saw radical to fix #658


davidsansome:
Thanks for fixing this! I'll kick off a release this evening.

PR #674, opened 2023-12-27
Add extra practice methods: recent mistakes, recent lessons, burned items

### LINKED ISSUE #446: Feature Request: Offline self-study quizzes
Generalised version of #332 - it would be great if there was a way to initiate self-study quizzes, similar to the website userscript, and ideally while offline. (As as an aside - I spend a lot of hours on the train with spotty network, so to me this would be an absolute killer feature, that I'd happily fork out 💰 for).

### LINKED ISSUE #627: Feature Request: Support for Extra Study
WaniKani introduced the Extra Study feature in March 2022. It would be handy to have access to this feature within the Tsurukame app if possible.

https://knowledge.wanikani.com/getting-started/extra-study/

### LINKED ISSUE #672: Feature Request: Recent Mistakes
This feature is available alongside reviews and lessons on the site. It would be nice to have it in the app as well.

Deadpikle:
This PR adds three new practice features to the app: review recent mistakes, review recent lessons, and review burned items.

### Practice: Review Recent Mistakes

Works like the website. You can review mistakes from the last 24 hours. Readings and meanings are grouped like on the website. This PR does not add the "redo lessons" feature like the website has.

Because [the API does not currently let you get review history](https://community.wanikani.com/t/api-changes-get-all-reviews/61617), however, this is only a local cache of recent mistakes. In theory, if/when that API comes back, this functionality will need to be updated/enhanced.

UI to get here is on the main screen and in the "Upcoming Reviews" section.

### Practice: Review Recent Lessons

Works like the website. You can review your recent lessons, review-style. Readings and meanings are grouped like on the website. The [definition of a "recent" lesson is found here](https://knowledge.wanikani.com/getting-started/extra-study/): 

> Any item you’ve learned in lessons and haven’t Guru’d in your reviews will be part of Recent Lessons mode. Once you complete your reviews and the item moves to the Guru stage, this will no longer remain in Recent Lessons.

UI to get here is on the main screen and in the "Upcoming Reviews" section (and above the recent mistakes review).

### Practice: Review Burned Items

Works like the website. You can review all your burned items review-style. Readings and meanings are grouped like on the website. I believe the WK website lets you practice all your burned items but doesn't repeat in between sessions until you have reviewed them all again. Right now, the app just throws all the burned items into a review session.

UI to get here is on the main screen all the way at the bottom under the "Burned" category.

### ~~Potential issue?~~

~~I think, for the practice sessions noted above, any changes to custom user data (custom meaning/reading explanation) aren't saved to the database. Am I reading the code correctly in that case? If so, this is probably a problem, since presumably a user might be using the explanation features when practicing recent mistakes. How should this best be solved? A refactor of `LocalCachingClient.sendProgress`? A new function to just save newer explanation data?~~ all is well, carry on.

### Fixes

* Fixes #672 
* Fixes #627 
* Fixes #446

### DIFF
--- a/ios/LocalCachingClient.swift
+++ b/ios/LocalCachingClient.swift
@@ -264,10 +264,12 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
 
     // Version 10. Only added to set the assignment is_kana_only_vocab field.
     "",
+    // Version 11. Added most_recent_mistake_time to allow for recent mistake reviews
+    "ALTER TABLE subject_progress ADD COLUMN last_mistake_time TIMESTAMP;",
   ]
 
   private let kInitialSchemaVersion = 8
-  private let kSchemaVersion = 10
+  private let kSchemaVersion = 11
 
   // Run when the user logs out. Clears everything in the database.
   private let kClearAllData = """
@@ -362,9 +364,30 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
     }
   }
 
+  func getAllRecentMistakeAssignments() -> [TKMAssignment] {
+    db.inDatabase { db in
+      getAllRecentMistakeAssignments(transaction: db)
+    }
+  }
+
   private func getAllAssignments(transaction db: FMDatabase) -> [TKMAssignment] {
     var ret = [TKMAssignment]()
     for cursor in db.query("SELECT pb FROM assignments") {
+      ret.append(cursor.proto(forColumnIndex: 0)!)
+    }
+    return ret
+  }
+
+  private func getAllRecentMistakeAssignments(transaction db: FMDatabase) -> [TKMAssignment] {
+    var ret = [TKMAssignment]()
+    let dayAgo = Calendar.current.date(byAdding: .hour, value: -24, to: Date())!
+    let formatter = DateFormatter()
+    formatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
+    for cursor in db.query("SELECT a.pb " +
+      "FROM subject_progress AS p " +
+      "LEFT JOIN assignments AS a " +
+      "ON p.id = a.subject_id " +
+      "WHERE last_mistake_time >= \"\(formatter.string(from: dayAgo))\"") {
       ret.append(cursor.proto(forColumnIndex: 0)!)
     }
     return ret
@@ -625,8 +648,24 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
     }
   }
 
+  func getRecentMistakesCount() -> Int {
+    let dayAgo = Calendar.current.date(byAdding: .hour, value: -24, to: Date())!
+    let formatter = DateFormatter()
+    formatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
+    return db.inDatabase { db in
+      let cursor = db.query("SELECT COUNT(*) FROM subject_progress " +
+        "WHERE srs_stage >= 1 AND last_mistake_time >= \"\(formatter.string(from: dayAgo))\"")
+      if cursor.next() {
+        return Int(cursor.int(forColumnIndex: 0))
+      }
+      return 0
+    }
+  }
+
   func sendProgress(_ progress: [TKMProgress]) -> Promise<Void> {
     db.inTransaction { db in
+      let formatter = DateFormatter()
+      formatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
       for p in progress {
         // Delete the assignment.
         db.mustExecuteUpdate("DELETE FROM assignments WHERE id = ?", args: [p.assignment.id])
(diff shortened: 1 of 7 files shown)


ios/LocalCachingClient.swift
```diff
@@ -373,6 +393,52 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
...
+    // and the item moves to the Guru stage, this will no longer remain in Recent Lessons.
+    // https://knowledge.wanikani.com/getting-started/extra-study/
+    var ret = [TKMAssignment]()
+    for cursor in db.query("SELECT a.pb " +
+      "FROM subject_progress AS p " +
+      "LEFT JOIN assignments AS a " +
+      "ON p.id = a.subject_id " +
+      "WHERE srs_stage >= \(SRSStage.apprentice1.rawValue) AND srs_stage <= \(SRSStage.apprentice4.rawValue)") {
+      let assignment = cursor.proto(forColumnIndex: 0) as TKMAssignment?
+      if assignment!.hasPassedAt == false {
```
davidsansome, line 435: `assignment` might be nil here - you should check it with an `if let assignment = ...` instead of asserting.

ios/LocalCachingClient.swift
```diff
@@ -627,8 +693,24 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
     }
   }
 
+  func getRecentMistakesCount() -> Int {
+    let dayAgo = Calendar.current.date(byAdding: .hour, value: -24, to: Date())!
+    let formatter = DateFormatter()
+    formatter.dateFormat = "yyyy-MM-dd HH:mm:ss"
+    return db.inDatabase { db in
+      let cursor = db.query("SELECT COUNT(*) FROM subject_progress " +
+        "WHERE srs_stage >= 1 AND last_mistake_time >= \"\(formatter.string(from: dayAgo))\"")
```
davidsansome, line 702: Use `\(SRSStage.apprentice1.rawValue)` here like above?

ios/MainViewController.swift
```diff
@@ -165,6 +165,8 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
 
     let lessons = services.localCachingClient.availableLessonCount
     let reviews = services.localCachingClient.availableReviewCount
+    let recentMistakes = services.localCachingClient.getRecentMistakesCount()
+    let recentLessons = services.localCachingClient.getAllRecentLessonAssignments()
```
davidsansome, line 169: It might be more efficient to add a separate `getAllRecentLessonAssignmentCount()` method that does a `SELECT COUNT(*)`. At the moment this is fetching a lot of data on every load of this page.

You could also consider caching the value for this one (with `@Cached`) since it doesn't depend on the current time. You'd need to invalidate it in `sendProgress`.

davidsansome:
> I think, for the practice sessions noted above, any changes to custom user data (custom meaning/reading explanation) aren't saved to the database. Am I reading the code correctly in that case?

It looks like `ReviewSession.addSynonym` does its thing even if `isPracticeSession` is true, so it should work okay. Did something not work when you tested it?

davidsansome (changes requested):

ios/MainViewController.swift
Deadpikle: I don't know that a `COUNT(*)` will work here as the only way to get an accurate count is to look at the `TKMAssignment.hasPassedAt` field, which isn't accessible in the database without loading the data. Maybe there's something I'm overlooking. In any case, I updated it to use a cache of the count value with the invalidation as you noted.

Deadpikle:
Thanks for the review. I've updated the code per your change requests and rebased on `master`. (Which greatly simplifies my commits since now the copyright year for most things has been updated, haha. Thanks for updating that.)

> > I think, for the practice sessions noted above, any changes to custom user data (custom meaning/reading explanation) aren't saved to the database. Am I reading the code correctly in that case?
> 
> It looks like `ReviewSession.addSynonym` does its thing even if `isPracticeSession` is true, so it should work okay. Did something not work when you tested it?

Ah. You're right. I retested. Works fine and as one would expect. Have edited the original post. My fault.

ios/MainViewController.swift
davidsansome: Ah yes I missed that. This looks fine then!

davidsansome:
This looks great, thanks for making these changes!

Deadpikle:
This is just a comment for documentation/future searches if it ever comes up: Right now, for recent mistakes, if you get it right again within 24 hrs, Tsurukame will no longer consider it a recent mistake. [This behavior matches WaniKani per this forum post announcing the feature.](https://community.wanikani.com/t/new-recent-mistakes-dashboard-component/62961)

> Q. A Recent Mistake I had in the past 24 hours disappeared, what happened?
> Most likely, you got that item correct in a Reviews session after getting it wrong in a previous Reviews session. If you get an item correct in a Reviews session before 24 hours have gone by, it disappears from the Recent Mistakes list.

So mistakes within 24 hrs that you then get correct in a subsequent review session should be and will be cleared from recent mistakes.

PR #675, opened 2023-12-27
Add ability to view list of items by SRS category

Deadpikle:
You now have the ability to view all apprentice/guru/etc. items at once. Useful especially for burned items.

Simulator Screenshot - iPhone SE (3rd generation) - 2023-12-27 at 14 42 11
Simulator Screenshot - iPhone SE (3rd generation) - 2023-12-27 at 14 43 18

### DIFF
--- a/ios/MainViewController.swift
+++ b/ios/MainViewController.swift
@@ -232,7 +232,8 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
     model.add(section: "All levels")
     for category in SRSStageCategory.apprentice ... SRSStageCategory.burned {
       let count = services.localCachingClient.srsCategoryCounts[category.rawValue]
-      let item = SRSStageCategoryItem(stageCategory: category, count: Int(count))
+      let item = SRSStageCategoryItem(stageCategory: category, count: Int(count),
+                                      accessoryType: count > 0 ? .disclosureIndicator : .none)
       if count > 0 {
         item.tapHandler = {
           self.selectedSrsStageCategory = category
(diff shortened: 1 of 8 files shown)


ios/MainViewController.swift
```diff
@@ -229,7 +231,14 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
     model.add(section: "All levels")
     for category in SRSStageCategory.apprentice ... SRSStageCategory.burned {
       let count = services.localCachingClient.srsCategoryCounts[category.rawValue]
-      model.add(SRSStageCategoryItem(stageCategory: category, count: Int(count)))
+      let item = SRSStageCategoryItem(stageCategory: category, count: Int(count))
```
davidsansome, line 234: Could you set `accessoryType: .disclosureIndicator` if `count > 0`, so it's obvious it can be tapped?

You'll probably have to add it to the `SRSStageCategoryItem` initialiser and pass it through.

ios/SubjectsByCategoryViewController.swift
```diff
@@ -0,0 +1,149 @@
...
+    self.showAnswers = showAnswers
+    setShowAnswers(showAnswers, animated: false)
+  }
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    navigationItem.title = category.description
+
+    let model = MutableTableModel(tableView: tableView)
+    model.add(section: "Radicals")
```
davidsansome, line 39: It looks a bit weird when these section headers are always added even if they're empty. Maybe you could only add the section headers if there are any subjects of that type? See `SubjectsRemainingViewController` for an example.

davidsansome (changes requested):
Thanks for this change, it looks great! Just a couple of comments!

Deadpikle:
Updated based on your review and rebased on latest `master`. ...not sure what happened with the disclosure indicators there; my screenshots have them but clearly they were gone in the code. Musta done something. Thanks for catching that and for reviewing.

PR #677, opened 2023-12-30
Fixes recent mistakes not sticking around after sync, cached #'s not being invalidated after logout/login

### LINKED ISSUE #676: Recent mistakes feature clears on updating from API
This is my fault on the new feature just merged in today. Recent mistakes work great offline until assignments are synced again from the internet and it clears the most recent mistake time in the local db. 

The `REPLACE INTO subject_progress` in `LocalCachingClient.fetchAssignments` is the suspected culprit. 

I am filing this as I will not have time for another day or so to fix. I'll take responsibility unless someone else gets to it first. Sorry about that.

Deadpikle:
My last PR, #674, had a small issue where recent mistakes were not sticking around in the database after an online assignment fetch due to `subject_progress` being wiped via a `REPLACE INTO` statement in `LocalCachingClient.fetchAssignments`. Instead of using `REPLACE INTO`, performs an `UPDATE` followed by an `INSERT` if the `UPDATE` did nothing.

Also fixes an issue I found where logging out of one account into another account without restarting the app didn't clear cached numbers in `LocalCachingClient`. Fixes this by invalidating items in `LocalCachingClient.clearAllData`. Also noticed that a force refresh would not clear the recent lesson count cached number from the web, so I also invalidated that on `sync`.

Closes #676

### DIFF
--- a/ios/LocalCachingClient.swift
+++ b/ios/LocalCachingClient.swift
@@ -1008,24 +1008,17 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
               try! assignment.serializedData(),
               assignment.subjectID,
             ])
-          // REPLACE INTO will set any non-specified column to NULL. This causes
-          // subject_progress.last_mistake_time to be cleared.
-          // In order to avoid deleting that data (and avoid reloading it just to
-          // use REPLACE INTO), perform an UPDATE first -- if a row was
-          // updated, all is well. If no update performed, we can perform an INSERT.
-          db.mustExecuteUpdate("UPDATE subject_progress SET level = ?, " +
-            "srs_stage = ?, subject_type = ? WHERE id = ?",
-            args: [assignment.level,
+          // perform an upsert since we don't want to lose any existing last_mistake_time
+          // data via REPLACE INTO.
+          db.mustExecuteUpdate("INSERT INTO subject_progress (id, level, " +
+            "srs_stage, subject_type, last_mistake_time) VALUES (?, ?, ?, ?, ?) " +
+            "ON CONFLICT (id) DO UPDATE SET level = excluded.level, " +
+            "srs_stage = excluded.srs_stage, subject_type = excluded.subject_type",
+            args: [assignment.subjectID, assignment.level,
                    assignment.srsStage.rawValue,
-                   assignment.subjectType.rawValue,
-                   assignment.subjectID])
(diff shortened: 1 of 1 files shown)


ios/LocalCachingClient.swift
```diff
@@ -1008,11 +1008,24 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
...
             ])
-          db.mustExecuteUpdate("REPLACE INTO subject_progress (id, level, " +
-            "srs_stage, subject_type) VALUES (?, ?, ?, ?)",
-            args: [assignment.subjectID, assignment.level,
+          // REPLACE INTO will set any non-specified column to NULL. This causes
+          // subject_progress.last_mistake_time to be cleared.
+          // In order to avoid deleting that data (and avoid reloading it just to
+          // use REPLACE INTO), perform an UPDATE first -- if a row was
+          // updated, all is well. If no update performed, we can perform an INSERT.
+          db.mustExecuteUpdate("UPDATE subject_progress SET level = ?, " +
```
davidsansome, line 1016: I think you might be able to do all this all in one statement with an [upsert](https://www.sqlite.org/lang_upsert.html).

Something like:

```
INSERT INTO subject_progress (...) VALUES (...)
ON CONFLICT (id) DO UPDATE SET ...;
```

davidsansome (changes requested):
Thanks for catching this bug!

Deadpikle:
Done and tested on iOS 17. That should work nicely.

Note for the future if it ever comes up: SQLite allows for this functionality as of version `3.24.0`. iOS [changes SQLite versions over time](https://stackoverflow.com/a/35993775/3938401), but iOS 12 has version `3.24.0`. Fortunately, Tsurukame also has a minimum version of iOS 12, so it should be fine to implement it this way. 👍

(P.S. Happy almost New Year!)

davidsansome:
> Done and tested on iOS 17. That should work nicely.
> 
> Note for the future if it ever comes up: SQLite allows for this functionality as of version `3.24.0`. iOS [changes SQLite versions over time](https://stackoverflow.com/a/35993775/3938401), but iOS 12 has version `3.24.0`. Fortunately, Tsurukame also has a minimum version of iOS 12, so it should be fine to implement it this way. 👍
> 
> (P.S. Happy almost New Year!)

Sweet, thanks for checking!
Happy new year to you too!

Issue #635, opened 2023-05-23
Gravatar Image Not Loading


bachmacintosh:
For some reason, Tsurukame is no longer loading my Gravatar from the email I use to log into WaniKani.

# Reproduction Steps

1. Uninstall Tsurukame if it's currently installed, and reinstall latest TestFlight Build
2. Restart iPhone
3. Sign into WaniKani in the app; the email address I use to sign in is *formatted* as `string.string@example.con`, I can provide an actual email to test if needed, but would prefer to do so privately.

## Expected Outcome 

The top-left icon in the app's main screen shows my [Gravatar](https://s.gravatar.com/avatar/cf832913d2b7c5ced15cf298839f64ee?s=80)

## Actual Outcome 

The default image displays instead of the Gravatar

![image](https://github.com/davidsansome/tsurukame/assets/3395010/2e281ef3-118f-4f68-bc71-c29322b2e492)

Let me know if any more info is needed.

Thanks!

Deadpikle:
@bachmacintosh Seems fixed per #680 and #681. Please try the latest TestFlight build! 😄

PR #680, opened 2024-01-01
Fixes and tweaks for Gravatar use

Deadpikle:
This PR rolls several small fixes/tweaks into one PR, all related to Gravatar use:

* Fixed the text color in `EditableTextModeItem` in dark mode (the email input for Gravatar was dark text on dark background in dark mode)
* Fixed Gravatar email input allowing more than one line (adds `maximumNumberOfLines` to`EditableTextModelItem`, implemented via/controlled by delegate method and setting `textView.textContainer.maximumNumberOfLines`)
* Added SHA-256 string hashing extension and used this for Gravatar hashes as [per documentation here](https://docs.gravatar.com/general/hash/)
* Allow manual clearing of image cache as debug step if avatar loading isn't working. This...honestly I'm not sure is needed. However, I am facing an interesting problem where the simulator and a newly built app on physical device loads custom avatars OK, and yet the TestFlight version doesn't. I'm wondering if my poor internet here caused the default image to be cached somehow, and thought I'd try this as a first step to testing for a fix. (See #635 -- I can get this user's Gravatar to load successfully in the app using the new custom email feature, so I'm not sure why there are issues for them or for me.) 
  * I have tried to replicate this error and cannot replicate it reliably. I'm wondering if perhaps there is a delay of some kind on the Gravatar end, the user's avatar isn't found, and then the default value is cached. Then the user can't clear the cache to try again. This is also why I am attempting an update to using SHA-256 hashes in case the current MD5 method has any issues. Sort of shooting in the dark, here. Open to ideas on how to debug this one.


davidsansome (approved):
I had a poke around on a real device and couldn't reproduce it reliably either.

I doubt clearing the cache will help since the different URLs should be different cache keys, but I guess it's worth a try.
What about adding some logging statements to MainTitleView to log the image URL. You'd be able to see them in the console for TestFlight builds.

Deadpikle:
Removed the MD5 files as you're correct, that was the only place it was being used.

> I doubt clearing the cache will help since the different URLs should be different cache keys, but I guess it's worth a try.

Yeah, sort of a shot in the dark if things are still not working with the SHA256 change for new URLs. Went ahead and added a debug line and I can hook up my phone into my Mac when the new TestFlight build goes out to test.

Rebased on latest `master`.

PR #683, opened 2024-01-02
A few small UI fixes

### LINKED ISSUE #578: Text background is wrong color in dark mode on Mac
If you zoom in, the text itself has a darker color behind it, closer to black. Ideally it would be transparent, or the same color as the cell background.

I think maybe this happened before with light mode, and we fixed it by switching to OS-provided colors instead of hardcoded values?

Screenshot 2022-08-06 at 11 51 38 AM

Deadpikle:
* Actually show an error message to the user on API login failure instead of just "Invalid API token!"
* Force login view to be in light mode at all times. This one is subjective, I'll admit, but I'd like to argue that dark mode doesn't look good on the login screen, and I'm not sure making an entirely new UI for login for dark mode is worth the effort. See screenshots below. At the very least, this will make it look better for now until someone with more UI skills than I can improve it in dark mode.
* Fixes #578 by setting subject details background color to `UIColor.clear` rather than `TKMStyle.Color.cellBackground`. Why this is needed is unknown to me right now as the same color comes out of `TKMStyle.Color.cellBackground` on iOS vs macOS yet macOS needs the tweak. It doesn't hurt anything on iOS since the cell background is already set in the, well, cell, and yes, normal color changes/highlights (e.g. for readings/meanings) still work fine.

Light mode login:
Simulator Screenshot - iPhone SE (3rd generation) - 2024-01-02 at 11 57 36
Dark mode login:
Simulator Screenshot - iPhone SE (3rd generation) - 2024-01-02 at 11 57 17

### DIFF
--- a/ios/LoginViewController.swift
+++ b/ios/LoginViewController.swift
@@ -46,6 +46,10 @@ class LoginViewController: UIViewController, UITextFieldDelegate {
   override func viewDidLoad() {
     super.viewDidLoad()
 
+    if #available(iOS 13.0, *) {
+      overrideUserInterfaceStyle = .light
+    }
+
     TKMStyle.addShadowToView(signInLabel, offset: 0, opacity: 1, radius: 5)
     TKMStyle.addShadowToView(privacyPolicyLabel, offset: 0, opacity: 1, radius: 2)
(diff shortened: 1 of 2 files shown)


davidsansome:
Awesome, thank you!

I think forcing light mode on the login screen is fine. I'm sure that used to look okay in dark mode ages ago - something must've changed at some point!

PR #687, opened 2024-01-10
Fixes: Recent mistakes that were in review queue did not show up in recent mistakes review

Deadpikle:
If an item is a recent mistake and is in the recent mistakes section BUT also was in the currently available review queue, it wasn't showing up in your recent mistakes queue. (The # on the main menu was correct and accurate, but when you tapped to do your recent mistake reviews, the items in the recent mistakes list that you were actually going to practice didn't match [there was a smaller number of things to practice]). 

Fixes the bug so what is on the main menu matches what you actually get to review/practice for recent mistakes — even if it is a recent mistake and is in the review queue, it should still be in the recent mistakes queue until 24 hrs have elapsed or until you get it right in a review.

Have verified that this behavior matches the website.

### DIFF
--- a/ios/ReviewItem.swift
+++ b/ios/ReviewItem.swift
@@ -61,7 +61,7 @@ class ReviewItem: NSObject {
                                           localCachingClient: LocalCachingClient) -> [ReviewItem] {
     filterReadyItems(assignments: assignments,
                      localCachingClient: localCachingClient) { assignment -> Bool in
-      assignment.isReviewStage && assignment.availableAtDate.timeIntervalSinceNow > 0
+      assignment.isReviewStage
     }
   }
 


UInt2048:
Isn't this creating invalid reviews if you push the review button? The reviews button isn't supposed to be an extra study queue.

Deadpikle:
> Isn't this creating invalid reviews if you push the review button? The reviews button isn't supposed to be an extra study queue.

No. Where are you seeing that? This is specifically a fix that is only used for the recent mistakes table cell button (the function is `readyForRecentMistakesReview`, NOT `readyForReview`), which is a newer feature in the current betas. The normal review process doesn't apply here; although it uses the same overall logic, the recent mistakes feature does not save review information to the WK backend (note `isPracticeSession` in `ReviewViewController.setup()`.

PR #701, opened 2024-03-21
Feature: Add lesson picker

Deadpikle:
Allows user to choose which lessons they want to do next. Based on feature from WK website. Lesson picker only shows up if user has available lessons and hasn't hit their apprentice count. Separates lessons by level and lesson type (radical vs kanji vs vocab).

Made some minor tweaks to the `SubjectModelItem` class to support using those types of cells with checkmarks. Added a new `LessonPickerViewController`. Other than that, pretty standard minor changes (new segues, main menu table cell item, etc.).

Note: Does not currently perform checks to see if user has selected more lessons to do than (current apprentice count + number of lessons selected to do).

Simulator Screenshot - iPhone SE (3rd generation) - 2024-03-21 at 19 17 46

### DIFF
--- a/ios/LessonPickerViewController.swift
+++ b/ios/LessonPickerViewController.swift
@@ -1,0 +1,149 @@
+// Copyright 2024 David Sansome
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


ios/LessonPickerViewController.swift
```diff
@@ -0,0 +1,149 @@
...
+  func setup(services: TKMServices) {
+    self.services = services
+  }
+
+  struct ReviewsForLevel {
+    var radicals: [SubjectModelItem]
+    var kanji: [SubjectModelItem]
+    var vocabulary: [SubjectModelItem]
+    init() {
+      radicals = []
```
davidsansome, line 35: Can you initialise these above, like `var radicals: [SubjectModelItem] = []`, and remove this init method?

davidsansome (approved):
I really like this!

ios/LessonPickerViewController.swift
Deadpikle: Done. 👍

Deadpikle:
Added # of lessons picked to the `Begin` button so that you can see how many you've picked. (The WK website has the same.)

PR #712, opened 2024-05-09
Remember recent mistakes after full refresh

Deadpikle:
In the current version of the app, because recent mistakes are stored in the local db, performing a full refresh from the WK server causes those mistakes to be lost, which can be very annoying if you do this accidentally.

This PR fixes that by:

1. Before refresh, load recent mistakes in last 24 hrs from local db.
2. After refresh and db updated, re-insert the recent mistake data into the local db.

NOTE: This does rely on subject IDs remaining the same before/after refresh. If only this data was easily available through the API 😞 Ideally these sorts of things this app stores would get synced to iCloud and thus show up on both macOS and iOS if you were using both devices or something, I suppose...

### DIFF
--- a/ios/LocalCachingClient.swift
+++ b/ios/LocalCachingClient.swift
@@ -734,6 +734,24 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
     }
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


UInt2048:
I wonder if there's a way to get recent mistakes incorporating any new data the refresh provides, because people might have done reviews on another device

PR #720, opened 2024-06-17
Adds swipe-to-go-back capability to the Apprentice, Guru, etc. categories.

jIdle:
Very small change. Adds swipe-to-go-back capability to the Apprentice, Guru, Master, Enlightened, and Burned views via the SubjectsByCategoryViewController. This way it is in line with the "Show remaining" view (SubjectsRemainingViewController).


davidsansome:
Nice, thanks!

PR #722, opened 2024-06-21
Fix bug in main VC if level exceeds max granted by subscription

### LINKED ISSUE #719: Finishing Level 3 on a free account results in wrong graph display
I just hit the end of level 3 on my free account but still have some vocabulary and even one kanji left to learn.

WaniKani shows me as 'level 4' but still lets me finish off level 3. Tsurukame on the other hand shows me as 'level 4' in the top bar with a 'current level 2' graph and empty 'level 3' graph below.

See the screenshot for better understanding:
image

UInt2048:
user.currentLevel is capped at maxLevelGrantedBySubscription which causes #719 due to an implementation oversight

Fixes #719

### DIFF
--- a/ios/MainViewController.swift
+++ b/ios/MainViewController.swift
@@ -234,16 +234,17 @@ class MainViewController: UIViewController, LoginViewControllerDelegate,
     }
 
     if Settings.showPreviousLevelGraph, user.currentLevel > 1,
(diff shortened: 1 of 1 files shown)


davidsansome (commented):
Thanks for the fix!

Issue #745, opened 2024-08-15
Review "batch size" setting is confusing


zjgoodman:
Someone reported in #586 that the review batch size "doesn't work." A contributor posted [this comment](https://github.com/davidsansome/tsurukame/issues/586#issuecomment-1446375391) providing an explanation that the "batch size" doesn't actually refer to the amount of items in a review session but instead refers to the internal selection pool of items when back-to-back reviews are **disabled**.

While the extended description of the functionality is sensible, the name of this setting is causing confusion. See below for the settings view of lessons and reviews.

# Lessons
![image](https://github.com/user-attachments/assets/219adb98-66e7-40c4-92bb-ea504778e813)
# Reviews
![image](https://github.com/user-attachments/assets/62af51d5-46b7-4ec5-8418-9b1afc1cc9e8)

Because both settings are called "batch size," a user would expect both of these settings to have the same effect. When in reality they do not. For lessons, it controls the number of items that appear in a lesson session. But for reviews, it controls the internal selection pool size for non-back-to-back reviews only.

I would propose the following:
1. change the `set the review queue size` description of the "review batch size" setting to provide additional explanation of what it does
2. if back-to-back reviews are enabled, disable and/or hide the "review batch size" setting entirely
3. rename the "review batch size" setting to something else so that it does not have the same name as LESSON "batch size" despite having different behavior

zjgoodman:
I am interested in contributing the fixes for this item

UInt2048:
@zjgoodman This is minor enough that you can probably get a PR merged. If you choose to rename it, I might suggest renaming it to "wrap-up queue size" in the user facing name. Avoid changing the internal name of the setting so that you don't affect people's already established preferences.

PR #746, opened 2024-08-15
docs(settings):  add extended description for "Review Batch Size"

zjgoodman:
This is to address issue #745

adds extended description to the "review batch size" property to attempt to clear up confusion about what the setting actually does

the description I wrote is:

> The \(name) setting is ONLY used when "back-to-back" reviews are disabled.
>
> When "back-to-back" reviews are disabled, you might be asked to review the meaning of an item and then later after reviewing some other items, be asked to review the reading of that item.
> The \(name) setting controls the maximum number of different review items you can encounter between reviewing the reading and meaning for a given item."

image

### DIFF
--- a/ios/ReviewSettingsViewController.swift
+++ b/ios/ReviewSettingsViewController.swift
@@ -358,9 +358,15 @@ func makeFontSizeViewController() -> UIViewController {
 }
 
 func makeReviewBatchSizeViewController() -> UIViewController {
+  let name = "Review Batch Size"
+  let description = "The \(name) setting is ONLY used when \"back-to-back\" reviews are disabled.\n\n" +
+    "When \"back-to-back\" reviews are disabled, you might be asked to review the meaning of an item and then " + 
+    "later after reviewing some other items, be asked to review the reading of that item. \nThe \(name) setting " + 
+    "controls the number of different review items you can encounter between reviewing the reading and " + 
+    "meaning for a given item."
   let vc = SettingChoiceListViewController(setting: Settings.$reviewBatchSize,
-                                           title: "Review Batch Size",
-                                           helpText: "Set the review queue size.")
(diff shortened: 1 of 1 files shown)


zjgoodman:
@davidsansome @UInt2048 this PR is ready for your review 👌

zjgoodman:
@davidsansome @UInt2048 could one of y'all review this when you have a chance?

ios/ReviewSettingsViewController.swift
```diff
@@ -358,9 +358,15 @@ func makeFontSizeViewController() -> UIViewController {
 }
 
 func makeReviewBatchSizeViewController() -> UIViewController {
+  let name = "Review Batch Size"
+  let description = "The \(name) setting is ONLY used when \"back-to-back\" reviews are disabled.\n\n" +
+    "When \"back-to-back\" reviews are disabled, you might be asked to review the meaning of an item and then " + 
+    "later after reviewing some other items, be asked to review the reading of that item. \nThe \(name) setting " + 
+    "controls the maximum number of different review items you can encounter between reviewing the reading and " +
```
davidsansome, line 365: This isn't technically correct - it controls the size of the pool of items from which items are randomly selected for review. So if you set it to 5, the probability of seeing 6 or more different items between the reading and meaning is `(4/5)^6`.

Anyway, it doesn't really matter. I'd change this to just remove the word "maximum" - it still gets the point across while being a bit less specific.

davidsansome:
Thanks for this change! Sorry it took me so long to look at it!

ios/ReviewSettingsViewController.swift
zjgoodman: Thanks for the review. I removed the word `maximum` as requested

PR #759, opened 2024-09-23
refactor(settings): rename 'batch size' property to be more descriptive

### LINKED ISSUE #745: Review "batch size" setting is confusing

zjgoodman:
closes #745 

renames the 'review batch size' property to reflect the actual behavior that it controls so that it's more obvious to users what they're actually changing when they tweak this setting

I landed on `Reviews Between Meaning & Reading` which I feel is extremely descriptive, yet concise and clear. I don't see how you could misunderstand the behavior of this property now.

# Settings screen
image

# Settings detail screen
image

### DIFF
--- a/ios/ReviewSettingsViewController.swift
+++ b/ios/ReviewSettingsViewController.swift
@@ -16,6 +16,7 @@ import Foundation
 import UIKit
 
 class ReviewSettingsViewController: UITableViewController, TKMViewController {
+  static let NON_B2B_REVIEW_BATCH_SIZE_PROP_NAME = "Reviews Between Meaning & Reading"
   private var services: TKMServices!
   private var model: TableModel?
   private var groupMeaningReadingIndexPath: IndexPath?
@@ -65,7 +66,8 @@ class ReviewSettingsViewController: UITableViewController, TKMViewController {
                                              hidden:!Settings
                                                .groupMeaningReading)
     nonBackToBackBatchSizeIndexPath = model.add(BasicModelItem(style: .value1,
-                                                               title: "Batch size",
+                                                               title: ReviewSettingsViewController
+                                                                 .NON_B2B_REVIEW_BATCH_SIZE_PROP_NAME,
                                                                subtitle: "\(Settings.reviewBatchSize.description)",
                                                                accessoryType: .disclosureIndicator) {
                                                   [unowned self] in self.didTapReviewBatchSize()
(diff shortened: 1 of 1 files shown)


ios/ReviewSettingsViewController.swift
```diff
@@ -363,11 +365,15 @@ func makeFontSizeViewController() -> UIViewController {
...
-    "controls the number of different review items you can encounter between reviewing the reading and " + 
+  let name = ReviewSettingsViewController.NON_B2B_REVIEW_BATCH_SIZE_PROP_NAME
+  let description =
+    "The \"\(name)\" setting is ONLY used when \"back-to-back\" reviews are disabled.\n\n" +
+    "When \"back-to-back\" reviews are disabled, you might be asked to review the meaning of an item and then " +
+
+    "later after reviewing some other items, be asked to review the reading of that item. \nThe \"\(name)\" setting " +
+
+    "controls the number of different review items you can encounter between reviewing the reading and " +
+
```
zjgoodman, line 376: this formatting change came from building the app. it's been doing this over and over and I've been manually reverting in my MRs but I figure I should let the formatter do its job so I allowed it to come into this MR

zjgoodman:
@davidsansome this one is ready for your review

zjgoodman:
@davidsansome whenever you get a chance can you please take a look at this one? Thank you 🙏

zjgoodman:
@davidsansome lemme know if you have any questions about this one 👌

davidsansome:
Sorry for the seriously long delay on this one!

PR #760, opened 2024-09-24
Fix text drawing for context sentence blurring

### LINKED ISSUE #718: Some text is clipped on iPhone
In some example sentences, the text is truncated. Please see the last example sentence in the screenshot which ends with “the engine will”

IMG_1775

I’m using an iPhone 14 Pro Max with iOS 17.5.1

### LINKED ISSUE #755: Missing translation to context sentence
IMG_8076

The translation of the first context sentence for the vocabulary ”emoji” is missing, even though it is present at the Wanikani website.

Deadpikle:
Closes #718
Closes #755 

It seems as though something is different in the way text is drawn in `ContextSentenceModelItem` and `AttributedModeltem` text measurement. Starting with iOS 16, `UITextView` changed its behavior for text layout. Reverting that change (to TextKit 1 layout) causes the Japanese text to draw the same for edge cases, but then the drawing in `ContextSentenceModelItem` is different.

To make it draw the same, grab the attributed string from the NSTextView and draw that instead. (The difference between the two attributed strings seems to be the font, interestingly enough, unless there is another attribute that I did not notice - I just checked via a simple print(), which may have skipped something).

So, said another way: The drawing in `ContextSentenceModelItem` draws different than the actual text layout in `AttributedModeltem`. In 生まれる, the first context sentence draws as 1 line in `ContextSentenceModelItem`, but 2 lines in `AttributedModeltem` via the actual `UITextView` (and for whatever reason the text height measurement is also only doing 1 line for the Japanese, making us 1 line short). We can make the `UITextView` draw the same way by forcing it to use TextKit 1 layout, but then the blurred text in `ContextSentenceModelItem` is not right, so we draw that text using the same attributed string as the `UITextView`, and things appear happy, even though I'm not exactly happy with this fix. But, I'd rather have something working than not working, even if it's ugly, since I'm not sure how to go about fixing this in the best way.

This is not the greatest fix in the world, I think, but it works on iOS 17 (SE 3 Simulator) and iOS 18 (SE 3 Simulator), and that is _something_ at least!

Test cases:
分
一台
二台
生まれ
生まれる
写る
足りない
交通
仮に
命
芸者
絵文字

All these test cases were broken before and are working now.

Before:
Simulator Screenshot - iPhone SE (3rd generation) - 2024-09-24 at 19 30 18
Simulator Screenshot - iPhone SE (3rd generation) - 2024-09-24 at 19 30 21

Note the wrong placement of the English blur (as it is expecting 2 lines for the Japanese text, but the Japanese text only takes 1 line when drawn this way). But then, when you tap it, the `UITextView` is drawing with 2 lines for the Japanese. The English always needs 2 lines, but the text measuring only assumes 3 lines total instead of 4.

After:
Simulator Screenshot - iPhone SE (3rd generation) - 2024-09-24 at 19 31 08
Simulator Screenshot - iPhone SE (3rd generation) - 2024-09-24 at 19 31 10

### DIFF
--- a/ios/Tables/AttributedModelItem.swift
+++ b/ios/Tables/AttributedModelItem.swift
@@ -44,7 +44,12 @@ class AttributedModelCell: TableModelCell {
     selectionStyle = .none
     isUserInteractionEnabled = true
 
-    textView = UITextView(frame: bounds)
+    if #available(iOS 16.0, *) {
+      textView = UITextView(usingTextLayoutManager: false)
+      textView.frame = bounds
+    } else {
+      textView = UITextView(frame: bounds)
+    }
     textView.isEditable = false
     textView.isScrollEnabled = false
     textView.textContainerInset = .zero
@@ -76,7 +81,7 @@ class AttributedModelCell: TableModelCell {
 
     // Calculate the height of the text. We can't just use UITextView.sizeThatFits because it gets
     // the wrong answer for CJK text. The order here matters, see
-    // https://github.com/facebook/AsyncDisplayKit/issues/2894.
+    // https://github.com/facebook/AsyncDisplayKit/issues/2894
     let storage = NSTextStorage()
     let manager = NSLayoutManager()
     manager.usesFontLeading = false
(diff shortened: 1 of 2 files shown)


cschmidt0121:
Thanks for this. Would love to see it merged.

davidsansome:
If it works it works! Thank you!

PR #763, opened 2024-10-12
Feature: iCloud sync for recent mistakes

### LINKED ISSUE #761: Feature: Syncing recent mistakes
Recent mistakes is nice, but the last little thing to make it "work well" (while the reviews API is unavailable...) would be to sync recent mistakes to iCloud so that a user could move between devices (e.g. phone, computer) and still do reviews and review recent mistakes.

Ideally, this should be a fairly easy operation:

On user finishing reviews:

1. Download JSON of recent mistakes (if it exists)
2. Merge with recent mistakes of just-completed-session, tossing out anything older than 24 hrs as that doesn't matter
3. Update local DB with recent mistakes (incl. anything from the cloud)
4. Create new JSON recent mistakes file
5. Upload to iCloud

On sync operation:

1. Download JSON of recent mistakes (if it exists)
2. If it exists, update local DB with recent mistakes from the downloaded file, taking care not to override any newer recent mistakes that are known locally for the same subject

I...have never written iCloud code before. I expected a simple "upload this single file and let me grab it later" operation to be simple, but it was not and/or I did not find a simple wrapper library for iCloud/CloudKit. [This library would work well](https://github.com/drewmccormack/SwiftCloudDrive), but it only works for newer versions of iOS/Swift.

So until I have more time, or someone much more knowledgeable than me in this area comes along, here is where I made potential progress:

_Save JSON recent mistake to a file_

```swift
    func mistakesUrl() -> URL {
      let paths = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)
      return URL(fileURLWithPath: "\(paths[0])/mistakes.json")
    }

    var jsonData: Data!
    do {
      var mistakes = getRecentMistakeTimes() // defined in LocalCachingClient.swift
      var arrForJson = [String: String]()
      mistakes.forEach { item in
        arrForJson[String(item.key)] = item.value // must have string keys otherwise JSONSerialization complains
      }
      jsonData = try JSONSerialization.data(withJSONObject: arrForJson,
                                            options: JSONSerialization.WritingOptions())
      let jsonString = String(data: jsonData as Data,
                              encoding: String.Encoding(rawValue: NSUTF8StringEncoding)) // debug
      print(jsonString) // debug
    } catch let error as NSError {
      print("Array to JSON conversion failed: \(error.localizedDescription)")
    }
    let jsonFilePath = mistakesUrl()
    do {
      try jsonData.write(to: jsonFilePath)
      print("JSON data was written to the file successfully!")
    } catch let error as NSError {
      print("Couldn't write to file: \(error.localizedDescription)")
    }
```

I had some attempts at saving and loading records from iCloud based on `CKRecord`/`CKAsset`, but that doesn't seem like the thing to do for this sort of operation. [This SO post](https://stackoverflow.com/questions/45597152/how-do-i-save-an-audio-file-to-icloud-using-swift) or [this one](https://stackoverflow.com/questions/7851039/how-to-use-icloud-to-store-and-sync-app-files) may help with actually reading/writing the file.

In addition, for this to work fully, the app creator for Tsurukame may need to create an official iCloud container for this.

### LINKED ISSUE #776: Review Recent Mistakes not showing up on ipad
I can’t seem to get the “review recent mistakes” section to show up on my iPad.  I can see it on my iPhone and both are running 1.33.1050. iPad Pro M1, iOS 18.0.1.

Deadpikle:
Uses `NSUbiquitousKeyValueStore` ([link](https://developer.apple.com/documentation/foundation/nsubiquitouskeyvaluestore)) to sync data between devices.

`NSUbiquitousKeyValueStore` allows for a max of 1 MB of data. I did a quick Excel check. WaniKani has a little less than or around 10,000 items. With max 25-byte pieces of data (subject ID and recent mistake time), we can store nearly 42,000 pieces of data, which is more than enough. Given that a user will probably have less than a 1,000 recent mistakes at any time (if they're super behind AND mess up a bunch), we are fine just using KVS instead of syncing a JSON file as thought about in #761.

If we need more space in the KVS for some odd reason in the future, we can compress the data slightly to save bytes. Again, I doubt we will ever reach any significant capacity for this feature, so it's likely nothing to worry about.

I successfully used this to sync recent mistakes from one iOS Simulator to another.

The one odd scenario at first glance here is as follows:

1. Device A uploads new recent mistakes
2. Device B starts. 
3. Before device B can download recent mistake data, it thinks there are no mistake, calls `updateRecentMistakesInCloud`, and says there are no recent mistakes. (Note that setting a key will not trigger the notification for data changed [as outlined here in the docs](https://developer.apple.com/documentation/foundation/nsubiquitouskeyvaluestoredidchangeexternallynotification?language=objc).)
4. Intuition might say we just hit a race condition, and I suppose we sort of did, BUT iCloud seems to be smart and still calls the notification for the key store changing from device A with the updated data in the preferences object, so everything syncs OK, at least in my test. (Even if the cloud data overwrites local data, since we still have the local data in the db and merge everything together, it all works out regardless.)
5. Once we get updated data from the cloud, in case we uploaded "bad data" previously OR if there is newer data the cloud didn't know about, we again update the KVS data, completing a full sync.

Closes #761
Closes #776


Deadpikle:
Hi. I don't want to be a pest, but since it has been 4 months, are there any plans to review and merge this PR? It does not require any extra iCloud setup (e.g. iCloud containers), fixes a long-standing issue with recent mistakes that others have mentioned besides me (#776), and generally has very low code-overhead.

davidsansome:
Hey! Sorry for not responding to this one. I took a quick look back in October but it looked like it would need more attention than I had at the time, and I never remembered to get back to it.

I think I'm nervous about this PR for a couple of reasons:

1. The encoding (comma-separated list of pipe-separated values) feels quite fragile. I'll be hard to safely extend in the future, and I'd worry about pipes or commas in values not being escaped properly. I think it would be better to use JSON if we care about it being human readable, or protobuf binary encoding if it's important to save bytes.
2. The merging code is a bit convoluted - it's hard to read, which means it's hard for me to be sure it's correct. I'd like to see some unit tests that cover the nominal cases and a few edge cases.
3. There are also a bunch of ways we could clean up the code to make it more readable. For example:
    - the comment in lines 1216-1219 describes the logic inversely to the code that follows it on lines 1220-1225
    - the repeated casts and force-unwrapping of optionals, like `Int32(subjectID)!` everywhere
    - mixing the deserialisation in the same function as the merging logic
    - generally convoluted logic, for example `updateRecentMistakesFromCloud` calls `mergeCloudWithLocalMistakes`, then calls `updateRecentMistakesInCloud` which calls `mergeCloudWithLocalMistakes` again
    - different names for different things in different places, for example `recentMistakesData` and `recentMistakes`
    - relying on string comparisons to compare datetimes. I know it happens to work with the iso8601 encoding we're using, but it feels fragile

If I was writing this I would try to make a data structure (struct or class) for the recent mistake data (essentially just the `[Int32: String]` that exists now, ideally change it to a `[Int32: Date]`). Give it a static method that creates one from the cloud, an initialiser that creates one from the data loaded from the database, a method for merging two instances, and a method that writes it back to the cloud. You could do all that in a separate file that has unit tests. Then `LocalCachingClient` will need methods for loading the dictionary from the db (we have this already), and writing it back. It would just be creating two of these recent mistake objects (one from the db, another from the cloud), merging them, and writing the final object back to both the db and the cloud.

I'm generally supportive of this feature and I'd be happy to merge the PR if we can clean it up a bit!

Deadpikle:
Hi @davidsansome,

Thank you for your kind and patient review for some code that, as you noted, needed to be reworked. I appreciate that. :) (And please don't feel bad for forgetting to reply or anything like that. It happens to all of us. If I ever ping you when I shouldn't, please feel free to call me out on it. We all have lives outside of this repo!)

I have reworked the code in major ways to clean it up and re-tested to make sure the sync works. (It's a bit of a pain: 2 iOS simulators logged into the same account; make mistake on one device, make sure the sync happens, wait a bit for the background iCloud sync to happen, pop open other device and watch it step through the breakpoints and see the new recent mistake show up.)

* Uploading and downloading to/from cloud is now in a new class, `RecentMistakeHandler`. It takes care of uploading and downloading the `[Int32: Date]` information. The data is converted to JSON, then uploaded as a single string to iCloud. We should still be well under the 1 MB limit.
* The merge code has been greatly simplified, commented, and unit-tested (See the `RecentMistakesTest` file). It should be significantly easier to understand, now.
* `LocalCachingClient` code has been simplified accordingly. The biggest thing it has now is a func that takes the local reviews, merges it with the cloud reviews, and updates the local db.
* Additional comments for specific code calls have been written to clarify intent and purpose.
* I thought about making `RecentMistakeHandler` monitor the notifications and then send out its own notification of updates, but this added unnecessary complexity.

Thanks again. Let me know what you think.

ios/RecentMistakeHandler.swift
```diff
@@ -0,0 +1,102 @@
...
+    keyStore.synchronize()
+  }
+
+  func getCloudStorageKey() -> String {
+    (fileNamePrefix ?? "") + "tsurukame-mistakes"
+  }
+
+  func getCloudMistakes() -> [Int32: Date] {
+    let data = keyValueStore?.data(forKey: getCloudStorageKey())
+    if data != nil {
```
davidsansome, line 41: You can do `if let data = ... {` here to avoid the force-unwrapping of `data` below.

ios/RecentMistakeHandler.swift
```diff
@@ -0,0 +1,102 @@
...
+      do {
+        let decoded = try JSONSerialization.jsonObject(with: data!)
+        if let mistakes = decoded as? [String: String] {
+          // re-create dict to be [Int32: Date] from [String: String]
+          var outputDict = [Int32: Date]()
+          mistakes.forEach { el in
+            outputDict[Int32(el.key)!] = dateFormatter.date(from: el.value)
```
davidsansome, line 48: I think this'll crash if the key is ever not a valid Int32. How about: 

```
if let key = Int32(el.key), let value = dateformatter.date(from: el.value) {
  outputDict[key] = value
}
```

ios/RecentMistakeHandler.swift
```diff
@@ -0,0 +1,102 @@
...
+          }
+          return outputDict
+        }
+      } catch {
+        NSLog("Unable to deserialize mistakes from JSON")
```
davidsansome, line 53: I think this throws away the error message. How about: `print("Unable to deserialize mistakes from JSON: \(error)")`

ios/RecentMistakeHandler.swift
```diff
@@ -0,0 +1,102 @@
...
+      mistakes.forEach { el in
+        strKeyDict[String(el.key)] = dateFormatter.string(from: el.value)
+      }
+      let json = try JSONSerialization.data(withJSONObject: strKeyDict)
+      keyValueStore?.set(json, forKey: getCloudStorageKey())
+      // sync latest date as well so that other devices are sure to get an update
+      keyValueStore?.set(Date(), forKey: "lastSyncCall")
+      keyValueStore?.synchronize() // fails silently if no iCloud account
+    } catch {
+      NSLog("Unable to serialize mistakes to JSON")
```
davidsansome, line 74: Here too

ios/RecentMistakeHandler.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  // Merge 2 dictionaries together and clean up any unnecessary data in the array along the way
+  static func mergeMistakes(original: [Int32: Date], other: [Int32: Date]) -> [Int32: Date] {
+    var output = [Int32: Date]()
+    let dayAgo = Calendar.current.date(byAdding: .hour, value: -24, to: Date())!
+    [original, other].forEach { dict in
+      dict.forEach { mistake in
```
davidsansome, line 83: I think `dict.forEach { (subjectID, dateToAdd) in` might work

ios/RecentMistakeHandler.swift
```diff
@@ -0,0 +1,102 @@
...
+        let subjectID = mistake.key
+        let dateToAdd = mistake.value
+        if dateToAdd >= dayAgo {
+          if let currentMistakeDate = output[subjectID] {
+            // compare the existing item in the output with the "new" item from this dictionary.
+            // We want the newer date of the two dates because we want the user to have the
+            // max 24 hrs for the recent mistake available.
+            output[subjectID] = dateToAdd > currentMistakeDate ? dateToAdd :
```
davidsansome, line 91: Does `max(dateToAdd, currentMistakeDate)` work here?

ios/LocalCachingClient.swift
```diff
@@ -737,18 +760,18 @@ class LocalCachingClient: NSObject, SubjectLevelGetter {
...
-    var mistakeTimes = [Int32: String]()
+    var mistakeTimes = [Int32: Date]()
     db.inDatabase { db in
       for cursor in db.query("SELECT id, last_mistake_time " +
         "FROM subject_progress " +
         "WHERE last_mistake_time >= \"\(dateFormatter.string(from: dayAgo))\"") {
         let subjectId = cursor.int(forColumnIndex: 0)
         let lastMistakeTime = cursor.string(forColumnIndex: 1)
-        mistakeTimes[subjectId] = lastMistakeTime
+        mistakeTimes[subjectId] = dateFormatter.date(from: lastMistakeTime!)
```
davidsansome, line 774: What does `dateFormatter.date(from:)` return if it couldn't parse the string? Do we need to check if it's nil?

```
if let date = dateFormatter.date(...) {
  mistakeTimes[subjectId] = date
```

ios/RecentMistakeHandler.swift
Deadpikle: Looks like `max` takes two `Comparable` items, and `Date` is `Comparable`, so yes!

ios/LocalCachingClient.swift
Deadpikle: Yeah, `dateFormatter.date(from:)` can return `nil`, so this has been changed to: 

```swift
if let lastMistakeTime = cursor.string(forColumnIndex: 1),
           let date = dateFormatter.date(from: lastMistakeTime) {
          mistakeTimes[subjectId] = date
}
```

(Added the other `let` into the `if` so that the force unwrap wasn't necessary.)

Deadpikle:
Oh, one other note:

`ENABLE_USER_SCRIPT_SANDBOXING = NO;` was changed to let me run unit tests. (See https://stackoverflow.com/a/76626156/3938401). I'm not sure if this was an issue with my version of macOS or Xcode or what, but it was needed to get the unit tests running.

I'm going to go ahead and set that to `YES` again as it seems like `YES` would be the correct default (we don't want to decrease security here by default), but I wanted to note that here in case anyone has issues running tests in the future.

Issue #625, opened 2023-04-16
Allow background audio


nzufelt:
When audio is playing in app, background music is temporarily paused. This is a bit jarring for those of us working in loud environments! I believe it’s a small setting within the “play audio” code to allow other audio (e.g. background music) to play in the background while this is playing. I’ll see if I can find it and make a PR.

UInt2048:
Here is the line responsible:

https://github.com/davidsansome/tsurukame/blob/e93e361c209b28e1dbb23f7e5cd888cdf15aa3e0/ios/Audio.swift#L45

As you can see, background audio not using the spokenAudio session mode will continue but a reduced volume, and other audio will be temporarily paused. This is intentional.

UInt2048:
You can read the documentation here if you're interested: https://developer.apple.com/documentation/avfaudio/avaudiosession/categoryoptions

charlesprior:
This could be a setting though. I’m sure many others like myself would prefer not to have the audio ducked/paused. We could have two settings for spoken and non-spoken audio each with the option of pause, duck and do nothing. I’d be happy to implement this if @nzufelt doesn’t want to.

davidsansome:
Thanks for the PR Paul! I think the one setting to interrupt/not interrupt background audio is fine - any more granular settings than that are probably less useful, and might get confusing.

PR #775, opened 2024-11-12
Issue 625 Allow background audio

paul-brenner:
@davidsansome Long time fan of Tsurukame. 

I often have music or something playing in my headphones and feel reluctant to jump in for a quick review because I know that every time audio plays it will harshly interrupt. I saw that someone else had the same request in #625 so added a setting to allow the option. 

Sorry to just barge in with a PR, not sure the etiquette for contributing on this project.


davidsansome:
This looks great, thanks for the PR!

Issue #622, opened 2023-03-25
Level mismatch


plinsley:
Level completed but Show Remaining still showing items from Level-1. This feels like a recent change in last month or so

plinsley:
![73E456BF-C1C8-4EA5-970A-FE9BD0578E26](https://user-images.githubusercontent.com/26242029/227745476-115054c3-1481-4697-9938-7bebf638bc56.png)
![9B9335BD-5CF5-42ED-B021-96F5DC3F0D9F](https://user-images.githubusercontent.com/26242029/227745479-1b422c81-e563-45ea-8946-044aa6d75f19.png)

plinsley:
Version 1.26.958

UInt2048:
This is a feature (see #606). You can disable it in the in-app settings > Radicals, Kanji, & Vocabulary > Keep Current Level Graph and change it to off.iPhoneから送信2023/03/25 17:34、Peter L ***@***.***>のメール:﻿
Version 1.26.958

PR #793, opened 2025-01-07
Tweak header names for current/next level

Deadpikle:
The header names when showing your prior level graph have caused some confusion (see #622). This PR attempts to address those concerns by:

* Changing "Current Level (#)" (which is actually the prior level) to "Prior level being completed (#)"
* Changing "Next Level (#)" to the more intuitive "Current Level (#)", which will match the "Current Level" number everywhere else, including WaniKani's website
* Adjusting the settings text accordingly for clarity

**This is a text only change. Feedback and improvements to wording very welcome, but wanting to clean this up a bit.**

### DIFF
--- a/ios/MainWaniKaniTabViewController.swift
+++ b/ios/MainWaniKaniTabViewController.swift
@@ -156,7 +156,7 @@ class MainWaniKaniTabViewController: UITableViewController {
        !services.localCachingClient.hasCompletedPreviousLevel() {
       let previousLevel = Int(user.currentLevel) - 1
       model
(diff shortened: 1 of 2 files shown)


davidsansome:
Thanks for the PR! I think it's a good change generally.

I'm not sure about the "being completed" part - I had to think a little bit to understand what it was trying to say. How about just "Prior level" or "Previous level"?

Deadpikle:
That makes more sense, yes. Wording has been tweaked!

davidsansome:
Looks great, thanks!

PR #797, opened 2025-01-13
feat(reviews): batch size

### LINKED ISSUE #586: Bugfix: Batch Size don't work in Reviews
The Batch Size is working only in Lessons but don't work in Reviews

zjgoodman:
Issue #586 was opened as a "bug" because the review "batch size" property wasn't consistent with the lesson "batch size" property and as a result, created false expectations from users about what the app was supposed to do. Lesson batch size sets the number of items that appear in a lesson. Users expected review batch size to do the same thing for reviews, but in reality, controlled something else. #745 was implemented to alleviate the pain of this problem.

This PR implements a new feature that closes #586 because it provides the behavior that users were originally expecting from the review "batch size" property.

# Home screen (no change)
From my home screen you can see that I have 284 reviews to complete. This is existing behavior. No change
01 base-reviews

# Review screen (no change)
When I start doing reviews, I can see that the review queue at the top right shows how many reviews I have. It's that same 284 reviews from earlier. The app will continue doing reviews until I completely finish all 284.
06 unlimited

# Review settings screen
In the review settings screen there is a new property called "review items in batches" that toggles on/off. The default value is OFF, making this feature an opt-in feature.
02 new-setting

# Review settings screen (with setting enabled)
When the new setting is enabled, another setting appears called "batch size." This is named consistently with the "batch size" setting in lesson settings.
03 new-setting

# Review batch size
When you click on the batch size setting you will be prompted to select a batch size that you want. 
04 new

# Review screen (with batch size set)
Now if I go back to the review screen from earlier, I can see that the review queue size at the top right reflects the desired batch size. I still have 284 reviews in my entire queue, but this review session will only have me do 30 before the ending the session.
05 limited

### DIFF
--- a/ios/MainWaniKaniTabViewController.swift
+++ b/ios/MainWaniKaniTabViewController.swift
@@ -1,4 +1,4 @@
-// Copyright 2024 David Sansome
+// Copyright 2025 David Sansome
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -231,10 +231,14 @@ class MainWaniKaniTabViewController: UITableViewController {
     switch StoryboardSegue.Main(segue) {
     case .startReviews:
       let assignments = services.localCachingClient.getAllAssignments()
-      let items = ReviewItem.readyForReview(assignments: assignments,
+      var items = ReviewItem.readyForReview(assignments: assignments,
                                             localCachingClient: services.localCachingClient)
       if items.count == 0 {
         return
+      }
+
+      if Settings.reviewItemsLimitEnabled && items.count > Settings.reviewItemsLimit {
+        items = Array(items[0 ..< Int(Settings.reviewItemsLimit)])
       }
 
       let vc = segue.destination as! ReviewContainerViewController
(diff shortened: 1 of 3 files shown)


zjgoodman:
@davidsansome this one is ready for review. Thank you! 🙏

davidsansome:
Thank you very much for the detailed PR description!

I really like this change.

Is the PR missing a commit at the moment? This commit only adds the new setting, it doesn't actually use it to limit the number of reviews presented to the user.

zjgoodman:
👀 hmmm if so then that’s strange. Maybe I messed up the commit squashing. I’ll check on that. Thanks On Jan 13, 2025, at 11:26 PM, David Sansome ***@***.***> wrote:﻿
Thank you very much for the detailed PR description!
I really like this change.
Is the PR missing a commit at the moment? This commit only adds the new setting, it doesn't actually use it to limit the number of reviews presented to the user.

zjgoodman:
Ok ok yes you were absolutely right. I just forgot to commit that file omg 😆 excellent catch. this should be ready to merge now

PR #801, opened 2025-01-16
fix(reviews): sort queue before truncating

zjgoodman:
I accidentally introduced a small bug in #797 where the review order isn't getting respected.

Screenshot 2025-01-16 at 10 14 03 AM

The reason for this is because the review queue starts off random, and I truncated the review queue in `ios/MainWaniKaniTabViewController.swift`, which then gets passed to `ios/ReviewSession.swift` where sorting happens. So long story short, I was truncating the queue BEFORE it got sorted. I updated the code to truncate after sorting and now everything is working as expected.

### DIFF
--- a/ios/MainWaniKaniTabViewController.swift
+++ b/ios/MainWaniKaniTabViewController.swift
@@ -237,10 +237,6 @@ class MainWaniKaniTabViewController: UITableViewController {
         return
       }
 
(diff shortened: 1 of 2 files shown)


zjgoodman:
@davidsansome this one is ready for review. Thanks

PR #806, opened 2025-02-19
Fix: Show API errors when account in hibernation

### LINKED ISSUE #767: Accounts in hibernation show no errors
If accounts are in hibernation and you are logged into the app, it will just keep silently failing and show you no errors. This case needs to be handled on the app (both at login and after) and users need to be pointed to the website to get their account out of hibernation mode.

(Short bug report for now, sorry, but wanted to log it before I forgot.)

Deadpikle:
Currently, no error is shown when accounts are in hibernation. This PR shows errors on the main screen and on login (for both username/password and API token) when the account is in hibernation. I don't think we can handle this for the user, so this points the user to the WaniKani website to fix the problem.

Screen shown on website when in hibernation:
Screenshot 2025-02-19 at 11 44 41 AM

What the app shows (this is the login screen but same idea if already logged in):
Simulator Screenshot - iPhone SE (3rd generation) - 2025-02-19 at 12 05 09

Fixes #767

### DIFF
--- a/ios/LoginViewController.swift
+++ b/ios/LoginViewController.swift
@@ -120,12 +120,9 @@ class LoginViewController: UIViewController, UITextFieldDelegate {
 
         self.delegate?.loginComplete()
       }.catch { error in
-        if let wkError = error as? WaniKaniAPI.WaniKaniWebClientError {
-          if wkError == .accountHibernating {
-            self.showHibernatingError()
-          } else {
-            self.showLoginError(error.localizedDescription)
-          }
+        if let wkError = error as? WaniKaniAPI.WaniKaniWebClientError,
+           wkError == .accountHibernating {
+          self.showHibernatingError()
         } else {
           self.showLoginError(error.localizedDescription)
         }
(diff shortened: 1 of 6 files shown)


ios/LoginViewController.swift
```diff
@@ -120,7 +120,15 @@ class LoginViewController: UIViewController, UITextFieldDelegate {
 
         self.delegate?.loginComplete()
       }.catch { error in
-        self.showLoginError(error.localizedDescription)
+        if let wkError = error as? WaniKaniAPI.WaniKaniWebClientError {
```
davidsansome, line 123: Can you combine these two checks into the single if statement? Something like `if let wkError = ..., wkError == .accountHibernating {`. That way you can get rid of one of the `showLoginError` branches.

ios/LoginViewController.swift
```diff
@@ -132,7 +140,15 @@ class LoginViewController: UIViewController, UITextFieldDelegate {
         Settings.userEmailAddress = ""
         self.delegate?.loginComplete()
       }.catch { err in
-        self.showLoginError("Unable to login with API token! (\(err.localizedDescription))")
+        if let wkError = err as? WaniKaniAPIError {
```
davidsansome, line 143: Same here too

ios/LoginViewController.swift
Deadpikle: Yes, good suggestion. Done! Thanks.

PR #811, opened 2025-03-03
Fix bug: Typing in not-latest position in TKMKanaInput moves caret to end

### LINKED ISSUE #743: Bug: Japanese text input moves focus to end when inserting kana
**Description**
On the Japanese text input, when the user finishes inserting a kana, the cursor moves to the end of the input. If the user isn't paying attention, this can result in entering the answer incorrectly.

**Steps to reproduce**
1. Focus the Japanese text input
2. Type "kanji" - it is converted live to かんじ
3. Select か
4. Type "ka" - it is converted to か
5. The cursor moves to the end of the input

**Demo**

https://github.com/user-attachments/assets/a69a0520-15f6-499f-b348-73932c87b262

**Expected behaviour**
The cursor should remain after か but before んじ.

**Additional information**
I can reproduce this issue on both macOS and iOS. On macOS, I'm using version 1.32.2 (1064).

Deadpikle:
When doing reviews, when you are typing at the end of the text field, all is well. However, if you make a mistake and go back to fix it, as soon as the text is transformed (say `ku` to `く`), the cursor auto-shifts to the end due to the text replacement in `TKMKanaInput`.

This PR fixes that by fixing the selection after the text has been changed.

Some GIFs for demonstration. (Not perfect GIF recordings, but they should give you the idea.)

Before:
before

After:
after

Fixes #743

### DIFF
--- a/ios/TKMKanaInput.m
+++ b/ios/TKMKanaInput.m
@@ -403,10 +403,12 @@ NSString *TKMConvertKanaText(NSString *input) {
         [kConsonants characterIsMember:newChar] && [kConsonants characterIsMember:lastChar]) {
       NSString *replacementString =
           (lastCharWasUppercase || _alphabet == kTKMAlphabetKatakana) ? @"ッ" : @"っ";
+      UITextRange *beforeChangeRange = textField.selectedTextRange;
       textField.text =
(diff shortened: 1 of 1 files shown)


davidsansome:
Amazing, thank you!

PR #813, opened 2025-03-09
Fix red text highlighting issue in reviews

Deadpikle:
After getting a shake and red letters, sometimes the full text would change red after making further changes despite changing the full text in TKMInput. Removing typing attributes fixes that. (The attributes were somehow carrying over into the full string when replacing text, likely because of where the user's cursor was, so removing those typing attributes removes the attributes moving into the full string.)

A reliable way to replicate is by going to a review for 分ける, typing `わけry`, hitting enter, erasing `y`, then typing. All text turns red.

Done in ReviewViewController since that's where the highlighting is done in the first place.

See also: https://stackoverflow.com/questions/34689141

Before:
before

After:
after

(P.S. Test builds are failing to publish right now; is there some new apple agreement that needs to be agreed to or something? GitHub Actions isn't listing the error publicly.)

### DIFF
--- a/ios/ReviewViewController.swift
+++ b/ios/ReviewViewController.swift
@@ -978,11 +978,12 @@ class ReviewViewController: UIViewController, UITextFieldDelegate, SubjectDelega
     submitButton.isEnabled = Settings.allowSkippingReviews || Settings.ankiMode || !text.isEmpty
   }
 
-  func textField(_: UITextField, shouldChangeCharactersIn _: NSRange,
+  func textField(_ field: UITextField, shouldChangeCharactersIn _: NSRange,
                  replacementString _: String) -> Bool {
     DispatchQueue.main.async {
       self.answerFieldValueDidChange()
     }
+    field.typingAttributes = [:]
     return true
   }
 


davidsansome:
I noticed this too but didn't get a chance to look into it yet, thanks for fixing it!

> (P.S. Test builds are failing to publish right now; is there some new apple agreement that needs to be agreed to or something? GitHub Actions isn't listing the error publicly.)

It looks like the provisioning profile was missing the iCloud capability, which is required for the recent-mistake syncing change. I'm trying to fix it now!

davidsansome:
> It looks like the provisioning profile was missing the iCloud capability, which is required for the recent-mistake syncing change. I'm trying to fix it now!

Yeah that did the trick - it seems happy again now!

Deadpikle:
> > It looks like the provisioning profile was missing the iCloud capability, which is required for the recent-mistake syncing change. I'm trying to fix it now!
> 
> Yeah that did the trick - it seems happy again now!

Ah, thank you! Sorry about that — I thought you wouldn't have to do anything on your end, but apparently I was wrong. Maybe I ticked a box on my dev stuff at some point without thinking about it. Apologies, and thanks!

PR #824, opened 2025-03-31
fix(reviews): sort queue when building session

### LINKED ISSUE #823: Lesson picker doesn't give me all my selected lessons
If you use the lesson picker feature to select the items you want to learn, you are allowed to select seemingly as many as you want, but when you start the actual lesson session it will have way less than what you selected.

# Select the lessons you want
Image

# Click quiz
Image

# Where did my lessons go?
Image

zjgoodman:
see #823 for background. Sorting the review queue inside the review session component causes everything that uses the reviewSession component (lessons, burned items, leeches, etc) gets truncated when we really just want to truncate the review session only

instead of sorting inside the review session component, sort when initializing the component from the main component

this was caused by #801 

closes #823

# Review session is now 16 items in length, as expected
Screenshot 2025-03-31 at 12 26 12 PM

### DIFF
--- a/ios/MainWaniKaniTabViewController.swift
+++ b/ios/MainWaniKaniTabViewController.swift
@@ -261,10 +261,17 @@ class MainWaniKaniTabViewController: UITableViewController {
     switch StoryboardSegue.Main(segue) {
     case .startReviews:
       let assignments = services.localCachingClient.getAllAssignments()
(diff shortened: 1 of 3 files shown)


zjgoodman:
@davidsansome this is ready for review. Thanks!

davidsansome:
Thanks for catching this!

PR #825, opened 2025-04-01
allow skipping kanji readings

zjgoodman:
# Feature
This pull pull request adds a new feature to settings / reviews that allows you to skip kanji readings in quizzes. **This impacts both reviews and lessons.**

# Background
I personally do not find a lot of value to learning the kanji readings through SRS because a given kanji can have lots of different readings depending on the word it's in.

One such instance for example is `日`, which has different readings in every single word below:
```
日本
七日
休日
日曜日
朝日
```

If you enable the `show all kanji readings` feature, Tsurukame will show you all readings for that kanji
image

However, Wanikani still only accepts `にち` as the "valid" reading for this kanji.

This is kind of an extreme example, since many kanjis really do only have one or two readings, but I have encountered several such examples in my studies to the point where I personally no longer find value in memorizing the "blessed" reading that wanikani wants me to learn. I personally prefer to indirectly learn readings through vocabulary words, and so for quite a while I have been ignoring kanji readings during quizzes and using the existing "allow cheating" feature of the app to skip the reading.

# The setting
The default value of this setting is `false` which means the default behavior of the app will be to quiz kanji readings unless explicitly overridden by the user.

Screenshot 2025-04-01 at 6 07 13 AM

# The behavior
When the setting is enabled, kanji will not ask for readings during quizzes. Vocabulary is unaffected.

demo3

### DIFF
--- a/ios/ReviewSession.swift
+++ b/ios/ReviewSession.swift
@@ -176,9 +176,12 @@ class ReviewSession {
     }
 
     // Remove it from the active queue if that was the last part.
+    let shouldIgnoreReading: Bool = activeSubject.hasKanji && Settings.skipKanjiReadings
+    let answeredReading: Bool = activeTask.answeredReading || activeSubject
+      .readings.isEmpty
     let isSubjectFinished =
       (activeTask.answeredMeaning || activeSubject.meanings.isEmpty) &&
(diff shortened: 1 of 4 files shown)


zjgoodman:
@davidsansome this is ready for review! Thanks!

PR #828, opened 2025-04-06
fix(practice): shuffle practice sessions

### LINKED ISSUE #827: Burned items reviews always showing the same items even if reviewed correctly.
Version 1.34.1062
IOS 18.4

When I review burned items, even by successfully reviewing them, everytime I stop and come back later, the same items appear again in the same order.
I might be doing something wrong, but I remember it wasn't behaving like that some time ago.
Here is a demonstration video:

https://github.com/user-attachments/assets/37cf84a9-7f79-4adf-a2aa-c56a207ec271

zjgoodman:
closes #827. see #827 for more info

TLDR the practice sessions are currently not being shuffled so when you do stuff like review burned items, you end up in the same order every single time. 

this PR shuffles practice sessions before initiating the review session so that you end up with fresh sessions each time

### DIFF
--- a/ios/MainWaniKaniTabViewController.swift
+++ b/ios/MainWaniKaniTabViewController.swift
@@ -266,7 +266,7 @@ class MainWaniKaniTabViewController: UITableViewController {
       if items.count == 0 {
         return
       }
-        
+
       items = sortReviewItems(items: items, services: services)
 
       if Settings.reviewItemsLimitEnabled && items.count > Settings.reviewItemsLimit {
@@ -281,7 +281,7 @@ class MainWaniKaniTabViewController: UITableViewController {
       let assignments = services.localCachingClient.getAllRecentMistakeAssignments()
       let items = ReviewItem.readyForRecentMistakesReview(assignments: assignments,
                                                           localCachingClient: services
-                                                            .localCachingClient)
+                                                            .localCachingClient).shuffled()
       if items.count == 0 {
         return
       }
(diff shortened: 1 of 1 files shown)


zjgoodman:
@davidsansome this is ready for review

ios/MainWaniKaniTabViewController.swift
```diff
@@ -279,9 +284,9 @@ class MainWaniKaniTabViewController: UITableViewController {
 
     case .startRecentMistakeReviews:
       let assignments = services.localCachingClient.getAllRecentMistakeAssignments()
-      let items = ReviewItem.readyForRecentMistakesReview(assignments: assignments,
+      let items = shuffle(ReviewItem.readyForRecentMistakesReview(assignments: assignments,
```
davidsansome, line 287: Does `ReviewItem.readyForRecentMistakesReview(...).shuffled()` work here?

If it does, we don't need the extra `shuffle` function!

davidsansome (changes requested):

ios/MainWaniKaniTabViewController.swift
zjgoodman: I believe that's a void function, which is probably why I ended up writing the new function but I will double check tomorrow. 👍

zjgoodman: Lmao nvm you are totally right. I will update my PR. Thanks for the callout

zjgoodman: Pushed correction. Ready for review 👌

PR #830, opened 2025-04-12
Re-integrate Settings.reviewBatchSize use

### LINKED ISSUE #829: Settings.reviewBatchSize is not used anymore
#824 removed the use of `Settings.reviewBatchSize`, which has changed the feel of reviews. Now review sessions with lots of reviews can have, well, any distance between reading and meaning, and saying "ask me later" no longer shifts the asked-later review to the end.

In `ReviewSession.swift`, `activeQueueSize = reviewQueue.count` is not correct. I am not sure why this change was made.

Deadpikle:
Reverts a small change in #824 to make sure the active queue size in reviews is correct.

@zjgoodman Could you help me understand why the change in #824 to remove the use of `Settings.reviewBatchSize` was made? The active queue size shouldn't be controlled by the number of reviews, I believe. I did a test, and the limitation for review items still works with the active queue size being set to `Settings.reviewBatchSize`.

Closes #829

### DIFF
--- a/ios/ReviewSession.swift
+++ b/ios/ReviewSession.swift
@@ -46,7 +46,7 @@ class ReviewSession {
       Settings.ankiModeCombineReadingMeaning) || self.isPracticeSession {
       activeQueueSize = 1
     } else {
-      activeQueueSize = reviewQueue.count
+      activeQueueSize = Int(Settings.reviewBatchSize)
     }
 
     refillActiveQueue()


zjgoodman:
Checking. Will report back asap

zjgoodman:
image

Ok yeah you're totally right! Idk wtf I was thinking when I made that change. That change was totally unnecessary and wasn't maintaining existing behavior. It should definitely be reverted. Good catch. Thanks!

Deadpikle:
> image
> 
> 
> 
> Ok yeah you're totally right! Idk wtf I was thinking when I made that change. That change was totally unnecessary and wasn't maintaining existing behavior. It should definitely be reverted. Good catch. Thanks!
> 
> 

Great. Thanks! 😀

davidsansome:
Thanks for noticing this!

PR #832, opened 2025-04-24
Add reader with audio playback to context sentences.

omarcande:
I'm not an expert iOS developer, but I wanted to add a reader that would playback audio for the context sentences. I feel this really helps with remembering the kanji pronunciation from the examples.

My approach was to reuse the existing pattern with the rightButton. I had to add NSObject as a parent object to be able to add the AVSpeechSynthesizerDelegate to ContextSentenceModelCell.

It probably needs a better implementation, but hoping this feature can be considered for a later release. Thanks!

### DIFF
--- a/ios/SubjectDetailsView.swift
+++ b/ios/SubjectDetailsView.swift
@@ -360,9 +360,11 @@ class SubjectDetailsView: UITableView, SubjectChipDelegate {
 
     let indexPath = model.add(section: "Context Sentences")
     for sentence in subject.vocabulary.sentences {
-      model.add(ContextSentenceModelItem(sentence, highlightSubject: subject,
-                                         defaultAttributes: defaultStringAttrs(),
-                                         fontSize: kFontSize))
+      let item = ContextSentenceModelItem(sentence, highlightSubject: subject,
+                                          defaultAttributes: defaultStringAttrs(),
+                                          fontSize: kFontSize)
+      item.initReader()
+      model.add(item)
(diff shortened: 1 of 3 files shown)


omarcande:
I tested this implementation further and noticed that the synthesizer made some pronunciation mistakes on basic words, such as pronouncing 日本人 as nihonhito instead of nihonjin. I did some further research and discovered an open source tool called Voicevox. This supports multiple voices and has far better pronunciation than the default synthesizer but it requires a host server for the engine. I decided to have the engine locally hosted and have an implementation that supports this tool. Not creating a PR for this, but here's my branch in case anyone is interested in the feature: https://github.com/omarcande/tsurukame/tree/feature/voicevox

omarcande:
@pocketlim Thanks for your feedback! I'm glad you like the idea. It has been of great use to me.

What I ended up doing is that I'm using the version that connects to my Voicevox server for personal use. However, since this is only available on my local network, I implemented a fallback to use the AVSynthesizer if the Voicevox server is unavailable. 

I'll take a look at the library you shared. The main problem with the AVSynthesizer is that the voice is robotic and monotonous. The actual pronunciation issues (such as pronouncing 日本人 as nihonhito) I have seen other text to speech also do. Seems most text to speech options still struggle with Japanese a bit.

I'm also currently working on adding a widget to display random review items. It's going well so far and I'll probably submit it as a PR just so that it's available for people to experiment with.

PR #835, opened 2025-05-05
feat: exclude items

zjgoodman:
### Short summary
this PR introduces a feature that allows users to **exclude** items that they don't care about, preventing them from appearing in their review queues

### Background
As I have progressed through wanikani, I have encountered various vocabulary words that I personally do not find value in studying, yet they appear in my review queue. they also contribute to my overall review count's number, which is currently quite high, which discourages me from doing reviews and also makes me feel like I'm not making much progress

Screenshot 2025-05-05 at 10 11 39 AM

A few examples of such words that I prefer not to study include:
```
金玉 testicles
ナポレオン3世 napoleon the 3rd
ハート形 heart shape
```

Additionally, there exists words that I do find value to learning, but there exist many duplicates of the words that are redundant and annoying to study. For example the counters:
```
~斤 bread loaf
一斤 1 bread loaf
二斤 2 bread loaves

~階 floor
一階 first floor
二階 second floor
四十二階 forty second floor
```

Another situation is words that are essentially the same thing but just stretched into 3 different items
```
消化する to digest
消化 digestion
不良消化 indigestion
```

# New behavior
## Main screen
At the bottom of the main screen under the "all levels" section, I added `excluded items`
main

## Excluded items view
When you click on the excluded items on the main screen, you are taken to the excluded items view.
excluded

## Item details view
If you click on a particular item in the excluded items view, you can see the usual item details screen, but now there's a handy slider to toggle whether or not the item should be excluded.
excluded-details

## From within a review
From within a review, users can exclude items the same way that they can add synonyms or mark their answer as correct.
review-exclude
Note that **Kanji and radicals cannot be excluded**. I decided to make it this way because:
1. This is a kanji learning app. Why would we exclude kanji?
2. If we allow users to exclude kanjis and/or radicals, they could prevent themselves from being able to level up
kanji

Putting it all together, here's what reviews look like:
review

# Persistence implementation
The way I implemented persistence might be controversial. I wanted to ensure that exclusions are saved across devices, so I decided to hijack the `reading note` field in wanikani to store my metadata in there

Excluded items will contain `#tsurukameExclude` in the note
wanikani

However, tsurukame will hide this text from you to prevent any confusion
tsurukame-no-note

For users who DO use this field, it's ok! You can still have notes for excluded items (although personally that seems like an unlikely scenario to me. if you excluded the item, why would you write notes for it?):
wanikani-note

The note also shows up in tsurukame
tsurukame-note

### DIFF
--- a/ios/LocalCachingClient.swift
+++ b/ios/LocalCachingClient.swift
@@ -49,6 +49,8 @@ private func postNotificationOnMainQueue(_ notification: Notification.Name) {
 }
 
 class LocalCachingClient: NSObject, SubjectLevelGetter {
+  let excludedText = "#tsurukameExclude"
+
   let client: WaniKaniAPIClient
   let reachability: Reachability
 
@@ -194,7 +196,7 @@ class LocalCachingClient: NSObject, SubjectLevelGetter {
     }
 
     let showKanaOnlyVocab = Settings.showKanaOnlyVocab
-    for assignment in getAllAssignments() {
+    for assignment in getNonExcludedAssignments() {
       // Don't count assignments with invalid subjects.  This includes assignments for levels higher
       // than the user's max subscription level.
       if !isValid(subjectId: assignment.subjectID) {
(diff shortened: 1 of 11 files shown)


zjgoodman:
@davidsansome this PR is ready for review. It's much bigger / complex than previous changes I've introduced but I tried to provide lots of explanation in the PR description for you. Lemme know if you have any comments / questions / concerns! Thanks!

zjgoodman:
@davidsansome when you get a chance, this PR is still waiting your review. I have a developer build of the app installed on my phone and I’ve been using this exclusion feature for the past two weeks and so far I haven’t encountered any problems with it. Very excited to get your feedback! Thanks!

ios/LocalCachingClient.swift
```diff
@@ -422,6 +424,60 @@ class LocalCachingClient: NSObject, SubjectLevelGetter {
...
+  }
+
+  func getMeaningNoteDisplay(studyMaterials: TKMStudyMaterials) -> String {
+    studyMaterials.meaningNote.replacingOccurrences(of: excludedText, with: "")
+  }
+
+  func setExcluded(studyMaterials: inout TKMStudyMaterials,
+                   shouldExclude: Bool) -> Promise<Void> {
+    if shouldExclude {
+      studyMaterials.meaningNote.append(excludedText)
```
davidsansome, line 438: I wonder whether we should add some whitespace (a newline or a space character) before the tag if the note is non-empty already? You'd have to remove it again (trim any trailing whitespace) in the else block below, and in `getMeaningNoteDisplay`.

ios/MainWaniKaniTabViewController.swift
```diff
@@ -230,6 +230,20 @@ class MainWaniKaniTabViewController: UITableViewController {
       }
     }
 
+    let excludedItems = BasicModelItem(style: .value1,
```
davidsansome, line 233: Maybe we could just not add this row if there are no excluded items? It'll just add clutter for people who don't use the feature.

davidsansome (commented):
Sorry for the slow reply! This looks good!

ios/MainWaniKaniTabViewController.swift
zjgoodman: Implemented! 👌

ios/LocalCachingClient.swift
zjgoodman: Implemented 👌

zjgoodman:
@davidsansome thanks for the review! I implemented your suggestions 👌

PR #840, opened 2025-07-06
Fix #451: Update visually similar kanji from API

### LINKED ISSUE #451: Update visually similar kanji automatically
Continued discussion from #448. In order to fix this issue, visually similar kanji need to be updated automatically in the same manner as subjects or assignments.

Visually similar kanji discrepancy

Deadpikle:
Allows for both updates from API and for manual visually similar kanji from JSON to keep previous handling of this feature while allowing for updates via the API to apply to the app. Prioritizes visually similar kanji from the API when sorting.

Allows removing restriction to skip showing visually similar kanji if said kanji are above current level with a setting. Defaults to off to maintain current/past behavior.

Required updating protobuf files.

Fixes #451


ios/WaniKaniAPI/Sources/WaniKaniAPI/Subject.swift
```diff
@@ -170,6 +170,9 @@ struct SubjectData: Codable {
       if let visuallySimilarKanji = kVisuallySimilarKanji[ret.japanese] {
```
davidsansome, line 170: I think I'd be happy removing this built-in list - it's cleaner to just rely on the data from the API. What do you think? Are the similar kanji in here better than Wanikani's?

davidsansome (approved):

Issue #865, opened 2025-12-22
Audio does not respect mute toggle when not using headphones


lazerwalker:
Hi!

Right now, if you have audio autoplay enabled in Tsurukame, but your iOS system-wide mute switch is on (either via Control Center or physical switch), it will still autoplay voice clips. 

It strikes me that a common iOS usecase is "I generally always want to autoplay WK audio clips, but I might be working through flashcards in public, in which case I temporarily don't want audio playing". The way iOS users would typically signal that intent is with the systemwide mute switch, rather than having to manually untoggle audio in Tsurukame / on a per-app basis.

This would be a one-line fix. Switching the AVAudioSession type from `.playback` to e.g. `.ambient` at https://github.com/davidsansome/tsurukame/blob/master/ios/Audio.swift#L45 would give the desired behavior (audio plays while either headphones are in or using the built-in speaker with the OS-level mute switch disabled). `.playback` is intended for media playback apps (e.g. YouTube, Spotify) where using the app without audio simply doesn't make any sense.

I'm more than happy to open a PR, but figured I'd open an Issue first since this is more of a UX question than a strict "bug", even if I personally think this is a cut-and-dry case of "what iOS users expect and what the Apple HIG dictates".

PR #866, opened 2025-12-22
Switch AVAudioSession category to .ambient

lazerwalker:
See #865.

Figured I'd open the PR anyway, to make it easier for someone to one-click merge in if you agree :D.

Per https://developer.apple.com/documentation/avfaudio/avaudiosession/category-swift.struct, `.ambient` is "The category for an app in which sound playback is nonprimary — that is, your app also works with the sound turned off." while `.playback` is "The category for playing recorded music or other sounds that are central to the successful use of your app.".

`.ambient` is specifically more appropriate here for the usecase I described — if someone is in public and has the system-level mute switch on, their expectation will be that Tsurukame does not autoplay voice clips.

### DIFF
--- a/ios/Audio.swift
+++ b/ios/Audio.swift
@@ -42,7 +42,7 @@ class Audio: NSObject {
     // Set the audio session category.
     let session = AVAudioSession.sharedInstance()
     try? session
-      .setCategory(.playback,
+      .setCategory(.ambient,
                    options: Settings
                      .interruptBackgroundAudio ? [
                        .duckOthers,


lazerwalker:
Thanks for the reply!

Ah, good catch on the background audio — yes, in my build, it seems like it just doesn't play if other background audio is playing. This is a plus for me personally (I find both the ducking and interrupt options to be distracting, and just don't use audio when listening to spoken text), which causes me to absolutely agree that this is too large a change to make without a better sense of broad user preferences.

That said, I wonder if you'd be open to making this a toggle, akin to the current `.interruptSpokenAudioAndMixWithOthers` and `.duckOthers` toggle? Could be something akin to "Respect system mute toggle" in the same two locations (settings menu and review slide-in menu). Thoughts?

> But on the other hand, I can imagine people wanting to use the mute switch to silence calls and alerts while concentrating on reviews - in which case they wouldn't want the switch to silence voice clips as well.

My mental model is that Do Not Disturb mode exists for that use case of "I don't want to receive other notifications while focusing on another app on my phone", while the mute toggle is specifically "if I'm using my phone in public, no app should make noise unless it's something exceedingly obviously like clicking play on a YouTube video". Tsurukame is one of the only apps I use regularly that does not respect the mute switch intent, and the fail state of "I forgot to explicitly turn off audio in the app just for this session" is quite bad if you're in a situation where stray audio is embarrassing. 

Apple's HIG text (https://developer.apple.com/design/human-interface-guidelines/playing-audio):
> People switch a device to silent when they want to avoid being interrupted by unexpected sounds like ringtones and incoming message tones. In this scenario, they also want to silence nonessential sounds, such as keyboard clicks, sound effects, game soundtracks, and other audible feedback. When a device is in silent mode, it plays only the audio that people explicitly initiate, like media playback, alarms, and audio/video messaging.

I think there's some level of ambiguity about whether the voice clips are "audible feedback" vs "audio that people explicitly initiate". But I'm taking the time to explain and emphasize my stance because while I think I agree this shouldn't be a unilateral forced change, I have to imagine there are other users who feel the same as me. In my mind, that's a strong argument for providing an option, the same way there is an option for background audio modes.

Curious for your thoughts!

PR #886, opened 2026-05-09
Fix card being skipped instead of retried after incorrect answer in groupMeaningReading mode

### LINKED ISSUE #883: Unwanted Skip when making a mistake
A bug skips cards with a mistake rather than letting me input the correct answer.

- when doing reviews
- I see the « vocabulary meaning » card, 
- if I make a mistake when responding
- I see the screen with the correct answer
- when I tap next to go back to the input screen and input the correct answer, it doesn’t show me the « vocabulary meaning » card again. Instead, my card is simply skipped and remains in my deck (it shows up later again)

This only started since the most recent update.
Version 1.35.1157
iPhone 16 Pro
 iOS 26.4.2

Sent a video and my cache to David via email for illustration

asantespi:
Fixes #883

## What was happening

In `groupMeaningReading` mode, answering a card incorrectly caused it to be skipped and pushed back into the deck rather than shown again for immediate retry. This was a regression introduced by PR #862 ("Improved Anki mode").

## Root cause

PR #862 added two things to `ReviewSession.swift` that conflict with each other in `groupMeaningReading` mode:

1. `returnDelay = 5` on any incorrect answer — intended to make the card reappear after 5 other items.
2. A "stay on current task" block in `nextTask()` that keeps the same card active for back-to-back meaning/reading — but only when `returnDelay == 0`.

The problem: in `groupMeaningReading` mode the active queue has size 1. When `returnDelay = 5` is set after a wrong answer, the card becomes ineligible, `eligibleIndices` is empty, and the code pulls a **new card** from `reviewQueue` to fill the slot. The original card is stranded and reappears much later.

Before PR #862, `nextTask()` simply used `arc4random_uniform(activeQueue.count)`. Since the queue only ever held 1 card in this mode, it always picked the same card — giving the implicit immediate-retry behaviour.

## Fix

Skip the `returnDelay` assignment in `groupMeaningReading` / practice mode. With `returnDelay` staying at `0`, the existing "stay on current task" logic in `nextTask()` correctly keeps the card active for retry:

```swift
case .Incorrect:
  tasksAnswered += 1
  // In groupMeaningReading/practice mode the queue has size 1; setting a delay
  // would evict the card and pull a new one in, breaking back-to-back retry.
  if !Settings.groupMeaningReading && !isPracticeSession {
    activeTask.returnDelay = 5
  }
```

Normal mode (without `groupMeaningReading`) is unaffected — it still gets the 5-item delay on incorrect answers.

### DIFF
--- a/ios/ReviewSession.swift
+++ b/ios/ReviewSession.swift
@@ -222,8 +222,11 @@ class ReviewSession {
 
     case .Incorrect:
       tasksAnswered += 1
-      // show this number of items before repeating this one:
-      activeTask.returnDelay = 5
+      // In groupMeaningReading/practice mode the queue has size 1; setting a delay
+      // would evict the card and pull a new one in, breaking back-to-back retry.
+      if !Settings.groupMeaningReading && !isPracticeSession {
+        activeTask.returnDelay = 5
+      }
 
     case .OverrideAnswerCorrect:
       tasksAnsweredCorrectly += 1


davidsansome:
Thank you!

Issue #881, opened 2026-05-02
Review item moving up and down when answering reviews


slimemold:
After answering a review, there’s this new (unintended) animation where it resizes the window and as a result moves the review item down and up. The review item moves because it’s vertically centered on the window, so the movement happens when the window is resized.

I don’t think it used to do this, but it’s extremely distracting. After a 60-item review, it made me want to throw my phone into the wall.

I’m on an iPhone 17 running iOS 26.4.2 (latest as of the time of this writing).

sixav:
I’m not sure if I’m facing the same exact problem but the effect seems to be similar, happened with the Version 1.35 update of the iOS app.
I have the setting “Switch to Japanese keyboard: Automatically switch to Japanese keyboard to type reading answers” turned on. I believe it used to switch the keyboard “in place” i.e. just like pressing the globe button. But now the keyboard quickly slides out of sight  and then pops back up, which is very visually distracting and results in the place of the current review item text moving too, like the original post mentions.

komodovaran:
Like this? It’s _quite_ annoying. Happened after most recent update. Also on latest version of iOS.

Goes away if you disable “Switch to Japanese keyboard” in the settings. Judging from the preferences icon it seems like this was a half baked update to Liquid Glass.

_Edit: Expired GIF link expired within a day_

slimemold:
Yup, that’s exactly what I’m talking about.

jlies87:
This exactly. It’s extremely annoying! I love this app, I’m on level 44 and use it daily but this keyboard slide is near app-breaking. It’s terribly visually distracting

tiffting:
https://github.com/user-attachments/assets/051858c8-28fd-4a51-86c5-305cfe5b4c52

jlies87:
If this can’t get fixed soon, is there a way to revert to a previous release? My reviews are mounting and I can’t get through them with this slide happening

Riplink:
Glad I’m not the only one having a problem with this. All The constant up and down started making me motion sick!

AndrewSouthpaw:
The keyboard also slides down on an incorrect answer, which is jarring. (Especially for me since I get so many wrong answers. 🥲) This regressed from the previous version. 

I think it's related?

https://github.com/user-attachments/assets/bc5b0d42-6e1c-48cc-8ea6-86b981165112

wrh7k68mwn-lgtm:
[...]

PR #887, opened 2026-05-11
Stabilize transition between en and jp keyboards

dshusta:
The newest release introduces a more disruptive dismiss-and-present animation when switching between en and jp keyboards, that was relatively subdued earlier. That answer field is flying all the way down and back up now, creating a really unstable UI experience. 

Changing to reloadInputViews makes the animation more direct keyboard-to-keyboard and keeps the UI in place.

| 1.34 tag in iOS 18.4 sim | Before on iOS 26 sim | PR on iOS 26 sim |
|--------|--------|--------|
| <video src="https://github.com/user-attachments/assets/5b1dcb3c-0216-4f84-b89c-786e2ca3b861" /> | <video src="https://github.com/user-attachments/assets/b975c6e4-7ac2-4d2b-97e5-5300e485fcdb" /> | <video src="https://github.com/user-attachments/assets/cea6241c-5223-4b96-9458-4ae5decb5f19" /> |

### DIFF
--- a/ios/AnswerTextField.swift
+++ b/ios/AnswerTextField.swift
@@ -30,6 +30,7 @@ class AnswerTextField: UITextField {
     didSet {
       if oldValue != useJapaneseKeyboard {
         if isFirstResponder {
+          // Reload the keyboard if we just changed its language.
           reloadInputViews()
         }
       }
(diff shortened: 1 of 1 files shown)


dshusta:
Honestly I'm also just flying through these reviews that much faster, i don't think I realized how much i was waiting on the keyboard switching animation

dshusta:
Addresses #881

davidsansome:
Thanks for fixing this! Sorry it took me such a long time to review.

PR #891, opened 2026-06-02
Fix keyboard dismiss and reappear

just-chillin:
Since iOS 26 I've had an issue where upon language change, the keyboard dismisses and then re-opens. This PR should fix that issue according to claude, as it reloads the keyboard without dismissing the view.

I tested it with the simulator, and while I wasn't able to reproduce the issue on a local build, the PR did not seem to negatively affect the keyboard.


Deadpikle:
This is the same fix as #887

davidsansome:
Thanks for the PR! I've merged #887 instead.

PR #898, opened 2026-08-22
Keep the keyboard up when a wrong answer is displayed

AndrewSouthpaw:
When an answer is marked incorrect, `markAnswer` disabled the answer field, which would cause the keyboard to slide off the screen. This was a disruptive flow, of the keyboard sliding in and out constantly. In earlier builds, the keyboard would stay present even with the answer marked wrong.

To be honest, these changes are all generated by AI, because I don't know Swift. I'm not sure if any of these changes are not idiomatic, and apologies if it misses the mark!

The same, answering a review, the keyboard in:

Screenshot 2026-08-22 at 9 38 24 AM

Before, the keyboard goes away:

Screenshot 2026-08-22 at 9 38 31 AM

After, the keyboard stays:

Screenshot 2026-08-22 at 9 31 34 AM

## AI slop details

Disabling a `UITextField` that is first responder makes it resign, so the keyboard slid off screen at the red answer and slid back on for the next subject. With a wrong answer every few reviews the keyboard is constantly moving, and the question label moves with it.

Instead of disabling the field, this keeps it focused and locks it: `isShowingWrongAnswer` makes `textField(_:shouldChangeCharactersIn:replacementString:)` reject edits, and the caret is hidden so the field doesn't look editable. The existing behaviour is otherwise unchanged - backspace clears the wrong answer so it can be typed again, Return or the forward button continues, and revealing the subject details still dismisses the keyboard, since the details view needs the space.

`TKMKanaInput` called its delegate's `textField:shouldChangeCharactersInRange:replacementString:` but discarded the result, so a `false` from the view controller had no effect. It now passes that through (guarded with `respondsToSelector:`, since the method is optional and the unit test stub doesn't implement it).

The Return key commands are skipped while the answer field has focus, because `textFieldShouldReturn` already handles Return there and both firing would skip a subject on a hardware keyboard.

Related: #881 / #887 fixed the other source of keyboard movement during reviews, the en/jp keyboard switch.

Tested in the simulator: wrong answers keep the keyboard in place, backspace clears for a retype, Return advances, and Anki mode plus "Show answer immediately" are unaffected.


AndrewSouthpaw:
@davidsansome btw idk if you get notifs for new PRs that open, so I'm sending you one ping. 😄 Sorry for the noise if you're busy with other stuff!

davidsansome:
Sorry for taking such a long time to review this! I wanted to test it myself before merging it, but then got distracted by other things and forgot!

It looks great. Thank you!

Issue #896, opened 2026-07-20
Latest release reverts to old (non-Anki style) review behaviour?


allanw:
In the latest version on iOS, it seems that if I answer a card incorrectly, I immediately see the same card again in the review pile.

I believe that prior to this version, the review algorithm would shuffle this recently failed card so that it appears later in the queue.

I think the latter behaviour makes more sense (it's closer to how Anki works, and it makes it too easy to just show the failed card again immediately).

Issue #888, opened 2026-05-12
No longer able to retype incorrect answers


alison0226-cmyk:
Previously if you got the answer wrong you would retype the correct one. This is a main feature I like so I can reinforce and see the meaning since I forgot. It also changes my reflex on answering completely now and I keep getting them wrong because I’m used to retyping the answer, but the app has moved to the next word.

AndrewSouthpaw:
FWIW I like this. When you'd retype the answer, you wouldn't be prompted to answer it again later. Now the card gets shuffled back in the stack by a little bit, which is the behavior of Anki and WaniKani.

alison0226-cmyk:
I did notice that after a few rounds and really like that change. Perhaps it could be a feature you turn on or off, to “require retyping the correct answer”. When using quizlet for a classroom vocabulary list I really enjoyed that as a feature since writing/typing is very helpful for me personally rather than simply seeing the word.

crun:
this behaviour change is personally unwelcome. when i make an error, i want to fix it immediately. now it results in cascading errors when i didn't notice the card had changed. revert back to the previous behaviour or at least make it a user option.

ClearOcean9:
Yes, it was so much better when you could immediately fix your mistake.  Immediate reinforcement.  The mistake adds the kanji to a later review anyway.   Also under the current less useful change, if you get the meaning correct and not the reading, the kanji get back into the shuffle but only the reading.  This is makes the flow very difficult because when a kanji appears I always type in the meaning then reading when a kanji appears, but not it only wants a reading so I make an additional error when I type the meaning.  I now have to carefully check whether meaning or reading is asked rather than being able to do it automatically so it slows things down a lot!  Please add the old system back as an options.  It is faster and provides immediate review

tarcisiotm:
I was about to open a new bug when I found this already opened.

Aside from all the points already raised, there is also the issue of not being able to let a review be processed as "wrong" at all, resulting in false corrects.

When I make a mistake, the card goes to the back of the stack but it doesn't remember it was already tried, almost as if we clicked "ask again later.

Version: 1.35.1157
Settings:

![image](https://github.com/user-attachments/assets/e5ee5e6f-c961-4e1b-b2cd-6a2fa75dce6e)

![image](https://github.com/user-attachments/assets/cfb47956-160b-4459-9daf-4ae6d88bb1eb)

Quick video:
[...]

PR #899, opened 2026-09-01
Make the delay before repeating an incorrect answer configurable

AndrewSouthpaw:
Adds a "Repeat incorrect answers" setting so people can choose whether an item they got wrong comes back immediately or after a few other items.

"Repeat incorrect answers" sits in the Order section of review settings and applies in every mode. "Immediately" asks the same item again straight away, so you can retype the correct answer. "After N items" shows N other items first.

The default is 5, which restores pre-#886 behavior, but I am fine with setting the default to be whatever you prefer.

 

### How the two behaviours got here

#862 made an incorrect answer set `returnDelay = 5`, so the item came back 5 items later instead of straight away.

In back-to-back and practice sessions the active queue has size 1, so a non-zero delay left nothing eligible and `nextTask()` pulled a new item in instead. #886 (fixing #883) turned the delay off entirely in those modes to restore immediate retry.

Both behaviours have advocates, and the tracker has an open issue for each direction: #888 wants to retype the answer immediately, and #896 wants the item shuffled back into the queue, as Anki and WaniKani do. A [comment on #888](https://github.com/davidsansome/tsurukame/issues/888#issuecomment-3413814294) suggested making it a setting rather than continuing to flip between the two, which is what this PR does.


AndrewSouthpaw:
Similar here, @davidsansome idk if you get notifs for new PRs that open, so I'm sending you one ping.

PR #902, opened 2026-09-13
Show the review summary when ending a review session early

AndrewSouthpaw:
Fixes the missing review summary when you end a session from the quick settings menu. Regression from #862, which added `|| !reviewVC.canWrapUp` to the early return that skips the confirmation dialog. That return also pops back to the main screen, so the results were lost.

To reproduce on master: answer a few reviews, finish the one you're on, then end the session from the menu.

## Test Plan
- [x] Ran a session on my phone, the summary shows when I end it from the menu.


davidsansome:
Looks good, thank you!

Issue #656, opened 2023-08-30
Incorrect accuracy percentage during reviews


Twiglet1022:
During reviews there is a percentage shown in the top right, next to number of completed reviews and number of remaining reviews.

I noticed that this percentage is counting the times where you input the correct answer after you have already gotten it wrong and revealed the right answer. This causes the percentage correct to be overstated.

For example, I get my first two reviews right so it is at 2/2 = 100% (the 2/2 is not shown in the UI). I then get my 3rd review wrong so it drops to 2/3 = 67%. The correct answer was revealed to me and I tap on the arrow to acknowledge my wrong answer. I’m asked to input the answer again so I put in the correct answer. Now the accuracy is 3/4 = 75%.

The behaviour I would expect is for the accuracy to remain at 2/3 = 67%. The app has effectively counted the same review twice, first when I got it wrong, and again when I completed it after having just seen the right answer.

The same happens if you get the same answer wrong multiple times. There’s never a reason to count a wrong answer more than once because the answer is revealed to you after you get it wrong so all you have do is type what you just saw. But particularly with readings, it’s easy to make typos on a phone screen so it happens to me often that I keep typing “hi” instead of “ji” or “tau” instead of “tsu” multiple times in a row.

I realise it’s just cosmetic and doesn’t actually affect the upcoming reviews or what gets synced up with WaniKani, but it would be nice for that percentage to be correct and match the one shown at the end of the review session.

UInt2048:
There are actually two accuracy percentages — the accuracy percentage during reviews and the accuracy percentage at the end in the review summary.

I believe Tsurukame behavior is consistent with the website if I remember correctly,

davidsansome:
Yep this is intentional to match the behaviour of the website

PR #903, opened 2026-09-13
Count the success rate per item so it matches the review summary

AndrewSouthpaw:
Changes the percentage at the top of the review screen to count per item, the same way the review summary does. One card with the reading right and the meaning wrong showed 50% at the top and 0% in the summary. An item counts against the rate as soon as either part is wrong, so the number drops on the wrong answer instead of waiting for the item to finish. `ReviewItem.answeredCorrectly` is the one definition of a correct item, used by both places.

## Test Plan
- [x] Ran a session on my phone, the top number matches the summary.


AndrewSouthpaw:
If #902 also lands, `tasksAnsweredCorrectly` has no readers left. Happy to delete it in whichever one you merge second.

Deadpikle:
Doesn't making review % and summary % make the app not match the website?

See also: https://github.com/davidsansome/tsurukame/issues/656

In other words, the percentages being different is intentional.

davidsansome:
Thank you for the PRs!

I agree with you the way it works now is unintuitive, but it was done this way to match what the website does. I don't think we should change that.

AndrewSouthpaw:
Sounds good! TIL that's how WaniKani does it.

